PR #15, opened 2021-04-07
Fix #14

### LINKED ISSUE #14: nixosConfiguration escape hatch
https://github.com/gytis-ivaskevicius/flake-utils-plus/blob/51cb739c9c9c2258bc70747eb7bc22975ae244bd/systemFlake.nix#L8

It is not entirely clear or self evident why `fup` should provide this service.

Sure, it is nice to have, but it is also only a `// nixosConfigurations` away in the top level flake.

/cc @Pacman99

blaggacao:


gytis-ivaskevicius (changes requested):
This was put in place to avoid overwriting `nixosConfigurations`. Now that I look at it - all extra arguments that are passed to the function probably should be overwritten on top of already built flake (if that makes sense). This would fix this issue and result in cleaner implementation.

This particular patch is rejected.

blaggacao:
Your suggestion sound good to me, so that would mean `nixosConfigurations` would go into `...`, did I understand you correctly?

blaggacao:
I just have noticed so, that this would need a recursive update. So we'd have to copy them over from lib.

gytis-ivaskevicius:
This is so very annoying since that's quite a bit of code duplication. I Will look into it, maybe ill find some sane solution with good old `//`. Also, this kinda makes me wonder - maybe this project is too complex not to depend on nixpkgs

Closing this PR since this patch is not going to get merged anyways.

blaggacao:
> maybe this project is too complex not to depend on nixpkgs

I think this restriction should not be dropped before discussing extensively with zimbatm. I think he has stories to tell. I'm looking if we can piggy-back on the selectedSystem somehow (or even just a default x64 selected system).

But in the end, the cleanest solution would probably to copy library function in-tree.

In the end, this is an upstream design issue with how `stdlib` and `pkgs.lib` relate.

PR #17, opened 2021-04-07
Allow config system customization

Pacman99:
This allows users to pass the config system they want a profile built with and the output it should go in.
So you could now do:
```
configProfiles.NixOS = {
  builder = inputs.darwin.lib.darwinSystem;
  output = "darwinConfigurations";
}
```
Also added `defaultConfigBuilder` and `defaultConfigOutput` to make this a bit easier for those that only use one system. And renamed `nixosProfiles` to `configProfiles` with the proper deprecation line. But I actually think this should be named `hosts`, and then we can go `defaultHostOutput`. Or maybe configs could work.

This will prevent flake-utils-plus making an opinion on what config system, but nixos support is probably a bit better. And this should allow for darwin and home manager support. This could also work for robotnix,  but that would require a bit more rework to `configurationBuilder`

### DIFF
--- a/examples/somewhat-realistic-flake.nix
+++ b/examples/somewhat-realistic-flake.nix
@@ -6,6 +6,7 @@
     unstable.url = github:nixos/nixpkgs/nixos-unstable;
     utils.url = github:gytis-ivaskevicius/flake-utils-plus;
 
+    nix-darwin.url = github:LnL7/nix-darwin;
     home-manager = {
       url = github:nix-community/home-manager/master;
       inputs.nixpkgs.follows = "nixpkgs";
@@ -13,7 +14,7 @@
   };
 
 
-  outputs = inputs@{ self, nixpkgs, unstable, utils, home-manager }:
+  outputs = inputs@{ self, nixpkgs, unstable, utils, nix-darwin, home-manager }:
     utils.lib.systemFlake {
 
       # `self` and `inputs` arguments are REQUIRED!!!!!!!!!!!!!!
@@ -56,6 +57,21 @@
         ];
       };
 
+      hosts."HostNameThree" = {
+        # This host will be exported under the flake's `darwinConfigurations` output
+        output = "darwinConfigurations";
+
+        # Build host with darwinSystem
+        builder = nix-darwin.lib.darwinSystem;
+
+        # This host uses `channels.unstable.{input,overlaysBuilder,config,patches}` attributes instead of `channels.nixpkgs.<...>`
+        channelName = "unstable";
+
+        # Host specific configuration. Same as `sharedModules`
+        modules = [
+          (import ./configurations/Morty.home.nix)
+        ];
+      };
 
 
 
--- a/examples/somewhat-realistic-flake.nix
+++ b/examples/somewhat-realistic-flake.nix
@@ -41,12 +41,12 @@
 
 
       # Profiles, gets parsed into `nixosConfigurations`
-      nixosHosts.HostnameOne.modules = [
+      hosts.HostnameOne.modules = [
         (import ./HostnameOneConfiguration.nix)
       ];
 
 
-      nixosHosts.HostnameTwo = {
+      hosts.HostnameTwo = {
         # This host uses `channels.unstable.{input,overlaysBuilder,config,patches}` attributes instead of `channels.nixpkgs.<...>`
         channelName = "unstable";
 
@@ -73,7 +73,7 @@
 
 
 
-      # Shared modules/configurations between `nixosHosts`
+      # Shared modules/configurations between `hosts`
       defaultHostsAttrs = {
         modules = [
           home-manager.nixosModules.home-manager
--- a/examples/somewhat-realistic-flake.nix
+++ b/examples/somewhat-realistic-flake.nix
@@ -67,7 +67,7 @@
         # This host uses `channels.unstable.{input,overlaysBuilder,config,patches}` attributes instead of `channels.nixpkgs.<...>`
         channelName = "unstable";
 
-        # Host specific configuration. Same as `sharedModules`
+        # Host specific configuration.
         modules = [
           (import ./configurations/Morty.home.nix)
         ];
@@ -90,7 +90,7 @@
 
 
       # Shared modules/configurations between `hosts`
-      defaultHostsAttrs = {
+      hostDefaults = {
         modules = [
           home-manager.nixosModules.home-manager
           # Sets sane `nix.*` defaults. Please refer to implementation/readme for more details.
@@ -107,7 +107,6 @@
 
     };
 }
-
 
 
 
--- a/systemFlake.nix
+++ b/systemFlake.nix
@@ -5,15 +5,16 @@
 , supportedSystems ? flake-utils-plus.lib.defaultSystems
 , inputs
 
-, nixosConfigurations ? { }
-, sharedExtraArgs ? { }
-, hostDefaults ? { }
-, nixosProfiles ? { } # will be deprecated soon, use nixosHosts, instead.
-, nixosHosts ? nixosProfiles
 , channels ? { }
 , channelsConfig ? { }
-, sharedModules ? [ ]
 , sharedOverlays ? [ ]
+
+, nixosConfigurations ? { } # deprecate soon, no longer used or works
+, hostDefaults ? { }
+, nixosProfiles ? { } # will be deprecated soon, use hosts, instead.
+, hosts ? nixosProfiles
+, sharedExtraArgs ? { }
+, sharedModules ? [ ]
 
 , packagesBuilder ? null
 , defaultPackageBuilder ? null
--- a/systemFlake.nix
+++ b/systemFlake.nix
@@ -29,28 +29,42 @@ let
   inherit (flake-utils-plus.lib) eachSystem;
 
   # ensure for that all expected, but no extra attrs are present
-  validateHost = {
-      channelName ? null
+  validateHost =
+    { channelName ? null
     , system ? null
-    , modules ? []
-    , extraArgs ? {}
-  }: { inherit channelName system modules extraArgs; };
+    , output ? null
+    , builder ? null
+    , modules ? [ ]
+    , extraArgs ? { }
+    }: { inherit channelName system output builder modules extraArgs; };
 
   mergeHosts = lhs: rhs:
-  let
-    # convoluted nix-kell, but clean(er) english-german below :-)
-    _ = x: op: y: op x y;
-    oder = lhs: rhs: if lhs != null then lhs else rhs;
+    let
+      # convoluted nix-kell, but clean(er) english-german below :-)
+      _ = x: op: y: op x y;
+      oder = lhs: rhs: if lhs != null then lhs else rhs;
 
-    rhs' = { channelName = _ rhs.channelName; system = _ rhs.system; };
-    lhs' = { channelName = _ lhs.channelName; system = _ lhs.system; };
-  in 
-  {
-    channelName = rhs'.channelName oder (lhs'.channelName oder "nixpkgs") ;
-    system = rhs'.system oder (lhs'.system oder defaultSystem) ; # replace deaultSystem with x86_64-linux
-    modules = rhs.modules ++ lhs.modules ++ sharedModules;
-    extraArgs = sharedExtraArgs // lhs.extraArgs // rhs.extraArgs;
-  };
+      rhs' = {
+        channelName = _ rhs.channelName;
+        system = _ rhs.system;
+        output = _ rhs.output;
+        builder = _ rhs.builder;
+      };
+      lhs' = {
+        channelName = _ lhs.channelName;
+        system = _ lhs.system;
+        output = _ lhs.output;
+        builder = _ lhs.builder;
+      };
+    in
+    rec {
+      channelName = rhs'.channelName oder (lhs'.channelName oder "nixpkgs");
+      system = rhs'.system oder (lhs'.system oder defaultSystem); # replace deaultSystem with x86_64-linux
+      output = rhs'.output oder (lhs'.output oder "nixosConfigurations");
+      builder = rhs'.builder oder (lhs'.builder oder channels.${channelName}.input.lib.nixosSystem);
+      modules = rhs.modules ++ lhs.modules ++ sharedModules;
+      extraArgs = sharedExtraArgs // lhs.extraArgs // rhs.extraArgs;
+    };
 
   optionalAttrs = check: value: if check then value else { };
 
@@ -58,7 +72,7 @@ let
     "defaultSystem" # TODO: deprecated, remove
     "sharedExtraArgs"
     "inputs"
-    "nixosHosts"
+    "hosts"
     "hostDefaults"
     "channels"
     "channelsConfig"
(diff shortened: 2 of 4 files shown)


systemFlake.nix
```diff
@@ -5,14 +5,19 @@
...
 , channelsConfig ? { }
 , sharedModules ? [ ]
 , sharedOverlays ? [ ]
 
+, nixosConfigurations ? { }
+  # nixosProfiles should be deprecated since profiles aren't nixos specific
+, nixosProfiles ? { }
+, configProfiles ? nixosProfiles
+, defaultConfigOutput ? "nixosConfigurations"
+, defaultConfigBuilder ? null
```
Pacman99, line 19: I think these should be renamed to hosts, but I'll let you decide first @gytis-ivaskevicius.

systemFlake.nix
```diff
@@ -48,34 +53,42 @@ let
...
 
-  genericConfigurationBuilder = hostname: profile: (
-    let selectedNixpkgs = getNixpkgs profile; in
+  configurationBuilder = hostname: profile: (
+    let
+      selectedNixpkgs = getNixpkgs profile;
+      defaultBuilder =
+        if defaultConfigBuilder == null
+        then inputs."${channelNameFromProfile profile}".lib.nixosSystem
+        else defaultConfigBuilder;
```
Pacman99, line 64: keep the nice nixos support, while still not sacrificing convenience for non-nixos.

systemFlake.nix
```diff
@@ -48,34 +53,42 @@ let
...
+        ]
+        ++ sharedModules
+        ++ (profile.modules or [ ]);
+        extraArgs = { inherit inputs; } // sharedExtraArgs // profile.extraArgs or { };
+      };
     }
   );
+
+  mapAttrs' = f: set:
+    builtins.listToAttrs (map (attr: f attr set.${attr}) (builtins.attrNames set));
```
Pacman99, line 91: This is a copy from upstream, I don't know if you want to put this somewhere else, but I'm fairly certain its necessary to get the `output` option for profiles to work.

systemFlake.nix
```diff
@@ -106,9 +119,5 @@ otherArguments
   // optional defaultAppBuilder { defaultApp = defaultAppBuilder pkgs; }
   // optional devShellBuilder { devShell = devShellBuilder pkgs; }
   // optional checksBuilder { checks = checksBuilder pkgs; }
-)
-
-  // {
-  nixosConfigurations = nixosConfigurations // (builtins.mapAttrs nixosConfigurationBuilder nixosProfiles);
-}
+) // (mapAttrs' configurationBuilder configProfiles)
```
Pacman99, line 122: Not necessarily against the `nixosConfigurations //` part, but I'm not sure the best way to integrate with this.

Pacman99 (commented):
Just a couple things to consider

systemFlake.nix
```diff
@@ -48,34 +53,42 @@ let
...
+        modules = [
+          ({ pkgs, lib, ... }: {
+            networking.hostName = hostname;
+
+            nixpkgs = {
+              inherit (selectedNixpkgs) overlays config system;
+            };
+
+            system.configurationRevision = lib.mkIf (self ? rev) self.rev;
+            nix.package = lib.mkDefault pkgs.nixUnstable;
```
Pacman99, line 80: I think these work on darwin, but would deny home-manager support. I have a method to only set options if they exist, but it adds a bit of complexity:
```nix
{
type = "merge";
contents = [
  (optional (options ? nixpkgs) { nixpkgs = ... })
];
}
```

Pacman99: This would make this PR **actually** config system agnostic.

Pacman99: And then to ensure pkgs always gets passed at the very least it could be:
`if (options ? pkgs) { nixpkgs = ... } else { _module.args.pkgs = ... }`

systemFlake.nix
```diff
@@ -48,34 +53,42 @@ let
...
-  genericConfigurationBuilder = hostname: profile: (
-    let selectedNixpkgs = getNixpkgs profile; in
+  configurationBuilder = hostname: profile: (
+    let
+      selectedNixpkgs = getNixpkgs profile;
+      defaultBuilder =
+        if defaultConfigBuilder == null
+        then inputs."${channelNameFromProfile profile}".lib.nixosSystem
+        else defaultConfigBuilder;
+      builder = profile.builder or defaultBuilder;
```
Pacman99, line 65: This is where the config system agnostic comes in. You can pass `darwinSystem` or `nixosSystem` or even home-manager configuration. The idea is just a function that builds the configuration.

systemFlake.nix
Pacman99: considering https://github.com/gytis-ivaskevicius/flake-utils-plus/pull/15#issuecomment-814995499, `nixosConfiguations` is not going to work anyway. I would say just drop the escape hatch entirely, flake-utils-plus is general enough, where its pretty unecessary.

systemFlake.nix
```diff
@@ -106,9 +119,6 @@ otherArguments
   // optional defaultAppBuilder { defaultApp = defaultAppBuilder pkgs; }
   // optional devShellBuilder { devShell = devShellBuilder pkgs; }
   // optional checksBuilder { checks = checksBuilder pkgs; }
-)
-
-  // {
-  nixosConfigurations = nixosConfigurations // (builtins.mapAttrs nixosConfigurationBuilder nixosProfiles);
-}
+) 
+  // (mapAttrs' configurationBuilder configProfiles)
```
blaggacao, line 123:
```suggestion
# produces attrset in the shape of
# { nixosConfigurations = {}; darwinConfigurations = {};  ... } 
# according to profile.output or defaultConfigOutput
// (mapAttrs' configurationBuilder configProfiles)
```

systemFlake.nix
```diff
@@ -5,14 +5,19 @@
...
 , channels ? { }
 , channelsConfig ? { }
 , sharedModules ? [ ]
 , sharedOverlays ? [ ]
 
+, nixosConfigurations ? { }
+  # nixosProfiles should be deprecated since profiles aren't nixos specific
+, nixosProfiles ? { }
+, configProfiles ? nixosProfiles
+, defaultConfigOutput ? "nixosConfigurations"
```
blaggacao, line 18:
```suggestion
, defaultConfigType ? "nixosConfigurations"
```

systemFlake.nix
blaggacao, line 19: This line is a bit treacherous, it implies that there is _no_ `defaultConfigBuilder`

systemFlake.nix
blaggacao: The whole `ouput` string is actually duplicating information that we should the builder to produce by itself.

I cannot think of a case where there is not a 1:1 mapping between builder and it's output key.

So this input can be spared from the public facing api.

systemFlake.nix
```diff
@@ -5,14 +5,19 @@
...
 
-, nixosConfigurations ? { }
 , sharedExtraArgs ? { }
-, nixosProfiles ? { }
 , channels ? { }
 , channelsConfig ? { }
 , sharedModules ? [ ]
 , sharedOverlays ? [ ]
 
+, nixosConfigurations ? { }
```
blaggacao, line 14:
```suggestion
```

This PR assumes resolution of #14 (see end of file)

systemFlake.nix
Pacman99: While that may be true, using the method I said, make the output of the builder '{ nixosConfigurations = }` is much harder to implement and adds complexity for the user trying to use it.
Its much easier to use to go
```nix
defaultConfigBuilder = inputs.darwin.lib.darwinSystem;
defaultConfigOutput = "darwinConfigurations;
```
And reading someone's flake.nix is much easier with this method.

Pacman99:
```nix
defaultConfigBuilder = args: { darwinConfigurations = inputs.darwin.lib.darwinSystem };
```
Is very complex for the user, for us, and the reader of flake.nix.

systemFlake.nix
Pacman99: I think `null` and possibly a comment then dealing with the default later is much cleaner. Trying to make a default here is going to get unnecessary complex.

We could just add a comment, that `null` is not actually the default.

Pacman99: dropped these lines now defaults are set in code. Prefer #19 for defaults as not to increase flake-utils-plus api surface.

systemFlake.nix
Pacman99: Keeping output vs builder separation but dropped defaults, prefer #19.

systemFlake.nix
Pacman99: You can read my comment there for reasoning, but I would say drop `nixosConfigurations` entirely. It goes against the idea that flake-utils-plus is a tool.

systemFlake.nix
```diff
@@ -106,9 +114,8 @@ otherArguments
...
-)
-
-  // {
-  nixosConfigurations = nixosConfigurations // (builtins.mapAttrs nixosConfigurationBuilder nixosProfiles);
-}
-
+) 
+# produces attrset in the shape of
+# { nixosConfigurations = {}; darwinConfigurations = {};  ... } 
+# according to profile.output or defaultConfigOutput
```
blaggacao, line 120:
```suggestion
# according to profile.output or the default `nixosConfigurations`
```

blaggacao (approved):
LGTM, factoring out the way defaults are handled to #19 (for discussion) was an excellent idea! Thanks!

systemFlake.nix
gytis-ivaskevicius: Maybe we should create a separate file with all the copy-pastes from nixpkgs? It could be just straight up added to flake lib argumentnt

systemFlake.nix
```diff
@@ -5,14 +5,17 @@
...
-, nixosProfiles ? { }
 , channels ? { }
 , channelsConfig ? { }
 , sharedModules ? [ ]
 , sharedOverlays ? [ ]
 
+, nixosConfigurations ? { }
+  # nixosProfiles should be deprecated since profiles aren't nixos specific
+, nixosProfiles ? { }
+, configProfiles ? nixosProfiles
```
gytis-ivaskevicius, line 17: Not a big fan of the name, maybe `hosts` or something like that should be considered.

gytis-ivaskevicius (approved):
I'd like to see some sort of example flake for all the changes that are being done. It does not need to be a good example or anything, just basic definitions are fine.

systemFlake.nix
Pacman99: I like hosts, but I'll probably submit the name change to a separate PR, as to discuss each change on its own.

systemFlake.nix
Pacman99: @gytis-ivaskevicius what about putting it directly in `flake.nix`? since in staging systemflake now has access to flake-utils-plus lib.

gytis-ivaskevicius: I was thinking about just dropping an import somewhere in here https://github.com/gytis-ivaskevicius/flake-utils-plus/blob/master/flake.nix#L26-L27 but adding it flake.nix directly isn't a bad idea, tho probably it does not scale well in the long run

Pacman99: Wait isn't the first link also `flake.nix`, and thats what I was talking about.

gytis-ivaskevicius: All i mean is seperation of sorts like 
```nix
      flake-utils.lib
      // (import ./copy-pasta.nix {} )
      // rec {
      ```

Comment (Pacman99) [reply systemFlake.nix:86]: ahh got it, I'll probably just add to flake.nix for now to avoid too many unrelated changes in this PR. But refactoring should definiitely happen

**Comment by Pacman99** [comment]
So I just switched to host in the top-level api and moved `mapAttrs'` to flake.nix as to keep this PR focused on its goal. I'll add an example next, and this should be ready to go.

File: systemFlake.nix [hunk 12]
```diff
@@ -26,19 +27,21 @@
...
-        inherit channelName system; 
-        modules = sharedModules ++ modules;
-        extraArgs = sharedExtraArgs // extraArgs;
-      };
+    }:
+    {
+      inherit channelName system output builder;
+      modules = sharedModules ++ modules;
+      extraArgs = sharedExtraArgs // extraArgs;
+    };
```
Pacman99, line 42: I realized defaults have to be set like this, otherwise `defaultHostAttrs` has no effect. Its a bit more verbose, but I think it looks pretty good.

Pacman99:
Added an un-tested nix-darwin example. Home-manager is possible, but requires writing a custom builder for it. 
I think this is ready, outside of the weird `nixosHosts` naming, which I would prefer to fix separately.

systemFlake.nix
```diff
@@ -5,16 +5,17 @@
...
-, sharedExtraArgs ? { }
-, defaultHostAttrs ? { }
-, nixosProfiles ? { } # will be deprecated soon, use nixosHosts, instead.
-, nixosHosts ? nixosProfiles
 , channels ? { }
 , channelsConfig ? { }
-, sharedModules ? [ ]
 , sharedOverlays ? [ ]
 
+, nixosConfigurations ? { } # deprecate soon, no longer used or works
```
Pacman99, line 12: This is my solution to the "escape hatch" issue

blaggacao (changes requested):
Two things, as far as I could currently see

flake.nix
```diff
@@ -33,6 +32,8 @@
           })
           paths;
 
+        mapAttrs' = f: set:
+          builtins.listToAttrs (map (attr: f attr set.${attr}) (builtins.attrNames set));
```
blaggacao, line 36: Nothing should be part of the public `fup` api, unless it adds serious `fup`-native value.

systemFlake.nix
blaggacao, line 42: I don't agree with the semantics here:

`default*` and `shared*` ought to be equivalent alternatives.

Note, that the word "default" does not mandate "override" for mergables (`[]` & `{}`).

flake.nix
blaggacao: Insofar, I guess I disagree with https://github.com/gytis-ivaskevicius/flake-utils-plus/pull/17#discussion_r609937344

blaggacao:
I once more conclude that our best option is to deprecate the top level api items that this can replace, for the following reasons:

- Don't risks to introduce subtle api differences ("default" vs "shared" vs "override")
- We avoid having to define and document precedence of competing declarations )of otherwise aliased options). Who has precedence? `defaultSystem` or `defaultHostsAttrs.system`? (To be clear: the point here is I do _not_ want to have to answer that question to keep the API reasonable).

flake.nix
Pacman99: So I just added mapAttrs' to systemFlake. Its either we export it or keep it in the file. I see no reason to use mapAttrs' outside of this use, so I think its ok to keep it in systemFlake.

Pacman99:
Just cleaned up, I added the nixosHosts -> hosts rename here, since this is the PR where that becomes true. But overall I just followed the new host args merging code to include output and builder.

examples/somewhat-realistic-flake.nix
```diff
@@ -73,7 +89,7 @@
 
 
 
-      # Shared modules/configurations between `nixosHosts`
+      # Shared modules/configurations between `hosts`
       defaultHostsAttrs = {
```
blaggacao, line 93:
```suggestion
      hostsDefaults = {
```

examples/somewhat-realistic-flake.nix
```diff
@@ -56,6 +57,21 @@
...
+        # This host will be exported under the flake's `darwinConfigurations` output
+        output = "darwinConfigurations";
+
+        # Build host with darwinSystem
+        builder = nix-darwin.lib.darwinSystem;
+
+        # This host uses `channels.unstable.{input,overlaysBuilder,config,patches}` attributes instead of `channels.nixpkgs.<...>`
+        channelName = "unstable";
+
+        # Host specific configuration. Same as `sharedModules`
```
blaggacao, line 70:
```suggestion
        # Host specific configuration.
```

blaggacao:
Kudos for this great improvement! It is really a very well targeted generalization.

PR #18, opened 2021-04-07
Only set options if they exist

Pacman99:
This makes flake-utils-plus even more general, while still keeping the sane defaults.

### DIFF
--- a/systemFlake.nix
+++ b/systemFlake.nix
@@ -40,6 +40,8 @@ let
 
   inherit (flake-utils-plus.lib) eachSystem;
 
+  optionalAttrs = check: value: if check then value else { };
+
   otherArguments = builtins.removeAttrs args [
     "defaultSystem"
     "sharedExtraArgs"
@@ -73,15 +75,28 @@ let
     {
       inherit (selectedNixpkgs) system;
       modules = [
-        ({ pkgs, lib, ... }: {
-          networking.hostName = hostname;
+        ({ pkgs, lib, options, ... }: {
+          # 'mkMerge` to separate out each part into its own module
+          _type = "merge";
+          contents = [
+            (optionalAttrs (options ? networking.hostName) { 
+              networking.hostName = hostname; 
+            })
 
-          nixpkgs = {
-            inherit (selectedNixpkgs) overlays config system;
-          };
+            (if options ? nixpkgs then {
+              nixpkgs = {
+                inherit (selectedNixpkgs) overlays config system;
+              };
+            } else { _module.args.pkgs = selectedNixpkgs; })
 
-          system.configurationRevision = lib.mkIf (self ? rev) self.rev;
-          nix.package = lib.mkDefault pkgs.nixUnstable;
+            (optionalAttrs (options ? system.configurationRevision) {
+              system.configurationRevision = lib.mkIf (self ? rev) self.rev;
+            })
+
+            (optionalAttrs (options ? nix.package) { 
+              nix.package = lib.mkDefault pkgs.nixUnstable;
+            })
+          ];
         })
       ]
       ++ profile.modules;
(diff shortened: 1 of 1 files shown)


systemFlake.nix
```diff
@@ -60,15 +62,25 @@ let
...
-          nixpkgs = {
-            inherit (selectedNixpkgs) overlays config system;
-          };
-
-          system.configurationRevision = lib.mkIf (self ? rev) self.rev;
-          nix.package = lib.mkDefault pkgs.nixUnstable;
+        ({ pkgs, lib, options, ... }: {
+          # 'mkMerge` to separate out each part into its own module
+          _type = "merge";
+          contents = [
```
Pacman99, line 68: "mkMerge" is necessary to prevent infinite recursion. You can't just merge the modules together.

Pacman99: Also this is just a manual `mkMerge` from  nixos module system, if that wasn't clear already.

blaggacao (approved):
LGTM.

The use case is a generalization of config builders, as discussed in #17

So if a specific builder (e.g. a `hmConfigBuilder`) does not know a specific option (eg. `nixpkgs`), then it can proceed without error.

This assumes that no builder deviates from `NixOS/nixpkgs` in the interpretation of its options namespace.

Pacman99:
> This assumes that no builder deviates from NixOS/nixpkgs in the interpretation of its options namespace.

True, I think this is a safe assumption for `nixpkgs` and `networking`, perhaps not `system`.

systemFlake.nix
```diff
@@ -60,15 +62,28 @@ let
...
+              networking.hostName = hostname; 
+            })
+
+            (if options ? nixpkgs then {
+              nixpkgs = {
+                inherit (selectedNixpkgs) overlays config system;
+              };
+            } else { _module.args.pkgs = selectedNixpkgs; })
+
+            (optionalAttrs (options ? system.configurationRevision) {
```
Pacman99, line 79: Apparently this is valid nix syntax

systemFlake.nix
```diff
@@ -60,15 +62,28 @@ let
...
+              system.configurationRevision = lib.mkIf (self ? rev) self.rev;
+            })
+
+            (optionalAttrs (options ? nix) { 
+              nix.package = lib.mkDefault pkgs.nixUnstable;
+            })
+          ];
```
gytis-ivaskevicius, line 86: What is the reasoning behind all these optionalAttrs? Is this meant to support non-nixpkgs based systems something?

Also, for example `options ? nix` does not check if `nix.package` exists - is this intended?

gytis-ivaskevicius (approved):

systemFlake.nix
Pacman99: So you can't use `mkIf` while checking if an option exists, I've done a fair bit of experimentation with this, while trying to make devos features optional. You also can't do `// optionalAttrs`, that causes infinite recursion. This way each option is technically considered its own module, which can be evaluated properly.
I can switch to actually check if the entire option exists. My reasoning was that if the `nix` namespace exists, then its likely for the sub options to exist, but I'll change to check for `nix.package` specifically.

related: https://github.com/divnix/devos/pull/226 and I got the `mkMerge` strategy from https://discourse.nixos.org/t/how-to-check-is-an-option-available/2919/2

Pacman99: Ohh right and reason, is that these options won't try to be set if they don't exist, so yeah systems like robotnix, home-manager, nix-darwin could benefit from this. While not hurting existing nixos users.

Pacman99:
I don't think examples here is possible, since it doesn't actually change any systemFlake usage. I'll add good example to #17 which should get darwin/home-manager with this merged.

I also switch the `options ?` lines to be more specific, so this isn't assuming any namespaces except for `nixpkgs` - which has to be assumed.

PR #19, opened 2021-04-07
Add argument for default host attributes

Pacman99:
Generic argument to add defaults to all hosts. This could cover modules, extraArgs, system, and channelName. I would say keep some things, ex `sharedModules`, just for convenience and backwards-compatibility, but this is overall useful to prevent making the api surface too large. Whenever adding things to hosts, you no longer need to make a relevant default argument, users can just use `defaultHostAttrs`.

Example:
```nix
defaultHostAttrs = {
  modules = ...;
  channeName = ...;
  system = ...;
};
```

### DIFF
--- a/systemFlake.nix
+++ b/systemFlake.nix
@@ -7,6 +7,7 @@
 
 , nixosConfigurations ? { }
 , sharedExtraArgs ? { }
+, defaultHostAttrs ? { }
 , nixosProfiles ? { } # will be deprecated soon, use nixosHosts, instead.
 , nixosHosts ? nixosProfiles
 , channels ? { }
@@ -34,6 +35,7 @@ let
     "sharedExtraArgs"
     "inputs"
     "nixosHosts"
+    "defaultHostAttrs"
     "channels"
     "channelsConfig"
     "self"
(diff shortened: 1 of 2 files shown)


systemFlake.nix
```diff
@@ -7,6 +7,7 @@
 
 , nixosConfigurations ? { }
 , sharedExtraArgs ? { }
+, defaultHostAttrs ? { }
```
blaggacao, line 10: I think it would be preferable if equivalence was surfaced here instead of buried within the implementation. Let's think if something along the lines is possible:

```suggestion
, defaultHostAttrs ? {
  modules = sharedModules;
  extraArgs = sharedExtraArgs;
  system = defaultSystem;
}
```

gytis-ivaskevicius (approved):

blaggacao (approved):
Perfect & clean solution!

Please feel entitled to merge, if it's ready.

blaggacao:
@gytis-ivaskevicius we where wondering if you wanted to preserve `shared*` top level flat api or deprecate it at some point in the future.

We can see how usage plays out and make that call whonce we get there.

gytis-ivaskevicius:
I'd like to preserve `shared*` the API. It is a very nice and convenient way to deal with system configuration. One of the things that might be nice to do is some sort of separation like `sharedDarwinModules` and `sharedNixOSModules` but I am not sure about that.

Related: https://github.com/divnix/devos/pull/232#discussion_r609448708

PR #20, opened 2021-04-07
Fix #16: nixosProfiles -> nixosHosts

### LINKED ISSUE #16: profiles
https://github.com/gytis-ivaskevicius/flake-utils-plus/blob/51cb739c9c9c2258bc70747eb7bc22975ae244bd/systemFlake.nix#L9

`nixosProfiles` do contain system profiles, in a sense, but more accurately they contain host configs.

This clashes with devos' usage of the term `profiles` which is reserved for a very special meaning:

- profiles cannot declare options, but only make use of them. (while anything that declares options are _modules_)
- this rule makes modules a lot more reusable among the community

With `fup` reserving the name, we might have a problem finding an alternative naming scheme. (and also it's somewhat established, already)

/cc @Pacman99

blaggacao:
nixosProfiles might conflict a little too much with downstream naming.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ Looking to add a kick-ass repl to your config? Create and import something along
       # Supported systems, used for packages, apps, devShell and multiple other definitions. Defaults to `flake-utils.lib.defaultSystems`
       supportedSystems = [ "aarch64-linux" "x86_64-linux" ];
 
-      # Default architecture to be used for `nixosProfiles` defaults to "x86_64-linux"
+      # Default architecture to be used for `nixosHosts` defaults to "x86_64-linux"
       defaultSystem = "aarch64-linux";
 
       # Channel definitions. `channels.<name>.{input,overlaysBuilder,config,patches}`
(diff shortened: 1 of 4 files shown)


blaggacao:
Merging this as an implementation of authorized https://github.com/gytis-ivaskevicius/flake-utils-plus/issues/16#issuecomment-814999215

Pacman99:
I would have preferred discussing hosts vs nixosHosts first. That would help for #17 and doing a rename in one swoop would be better.

Pacman99:
I think everything should be hosts. Its getting very confusing to juggle around two names that are extremely similar. I say just rename to hosts and use host/hosts everywhere in fup. Leave the name `profiles` reserved for the more standard nixpkgs definition, a system agnostic config set.

blaggacao:
`profile` &rarr; `config`? &mdash; they are mostly the contents of a traditional `/etc/nixos/configuration.nix`?

gytis-ivaskevicius:
fup definition of `profiles is basically a "computer configuration". Word `host` seems to work fine in this context

PR #21, opened 2021-04-08
fix: defaultHostAttrs merging

blaggacao:
###### Since we are not yet completely clear on details of this fix, I reverted the offending commit in `staging` for now and reapplied it here, so that `staging` is kept free of offending commits until we find a solution.
---
if both, defaultSystem and defaultHostAttrs.system are declared, which
should take precedence? And at any rate the other one becomes a no-op
api argument. Conditionally no-op api args are treacherous.

To avoid this sort of problems all args with an override semantic
need to be removed from top level. Currently this affects defaultSystem.

Note that since sharedModules and sharedExtraArgs have merge semantics,
they don't suffer this conflict, if both they are declared in both
places, at the very least non of them becomes a no-op.

---

Also validate API contracts (almost) properly (still no type checking).

### DIFF
--- a/systemFlake.nix
+++ b/systemFlake.nix
@@ -7,6 +7,7 @@
 
 , nixosConfigurations ? { }
 , sharedExtraArgs ? { }
+, defaultHostAttrs ? { }
 , nixosProfiles ? { } # will be deprecated soon, use nixosHosts, instead.
 , nixosHosts ? nixosProfiles
 , channels ? { }
@@ -24,18 +25,29 @@
 }@args:
 
 let
+  evalHostArgs =
+    { channelName ? "nixpkgs"
+    , modules ? []
+    , system ? defaultSystem
+    , extraArgs ? {}
+    , ...
+    }: defaultHostAttrs
+      // { 
+        inherit channelName system; 
+        modules = sharedModules ++ modules;
+        extraArgs = sharedExtraArgs // extraArgs;
+      };
+
   inherit (flake-utils-plus.lib) eachSystem;
 
   optionalAttrs = check: value: if check then value else { };
-
-  channelNameFromProfile = profile: profile.channelName or "nixpkgs";
-  systemFromProfile = profile: profile.system or defaultSystem;
 
   otherArguments = builtins.removeAttrs args [
     "defaultSystem"
     "sharedExtraArgs"
     "inputs"
     "nixosHosts"
+    "defaultHostAttrs"
     "channels"
     "channelsConfig"
     "self"
@@ -51,12 +63,12 @@ let
     "checksBuilder"
   ];
 
-  nixosConfigurationBuilder = hostname: profile: (
+  nixosConfigurationBuilder = hostname: profile: 
+    let hostAttrs = evalHostArgs profile; in
     # It would be nice to get `nixosSystem` reference from `selectedNixpkgs` but it is not possible at this moment
-    inputs."${channelNameFromProfile profile}".lib.nixosSystem (genericConfigurationBuilder hostname profile)
-  );
+    inputs."${hostAttrs.channelName}".lib.nixosSystem (genericConfigurationBuilder hostname hostAttrs);
 
-  getNixpkgs = profile: self.pkgs."${systemFromProfile profile}"."${channelNameFromProfile profile}";
+  getNixpkgs = profile: self.pkgs."${profile.system}"."${profile.channelName}";
 
   genericConfigurationBuilder = hostname: profile: (
     let selectedNixpkgs = getNixpkgs profile; in
@@ -87,9 +99,8 @@ let
           ];
         })
       ]
-      ++ sharedModules
-      ++ (profile.modules or [ ]);
-      extraArgs = { inherit inputs; } // sharedExtraArgs // profile.extraArgs or { };
+      ++ profile.modules;
+      extraArgs = { inherit inputs; } // profile.extraArgs;
     }
   );
 in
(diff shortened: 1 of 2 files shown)


systemFlake.nix
```diff
@@ -26,24 +26,29 @@
...
-    }: defaultHostAttrs
-      // { 
-        inherit channelName system; 
-        modules = sharedModules ++ modules;
-        extraArgs = sharedExtraArgs // extraArgs;
-      };
+    }: {
+      # hard coded fallbacks
+      channelName = "nixpkgs";
+      system = defaultSystem or "x86_64-linux";
```
blaggacao, line 37: `nix` experts: is `null` "falsy"?

systemFlake.nix
```diff
@@ -26,24 +26,28 @@
...
+    }
+    //  defaultHostAttrs
+    // {
+      inherit channelName system; 
+      modules = sharedModules ++ (if defaultHostAttrs ? modules then defaultHostAttrs.modules else []) ++ modules;
+      extraArgs = sharedExtraArgs // (if defaultHostAttrs ? extraArgs then defaultHostAttrs.extraArgs else {}) // extraArgs;
```
blaggacao, line 42: nix experts: a more nixonic way here would be?

systemFlake.nix
```diff
@@ -26,24 +26,28 @@
...
-    }: defaultHostAttrs
-      // { 
-        inherit channelName system; 
-        modules = sharedModules ++ modules;
-        extraArgs = sharedExtraArgs // extraArgs;
-      };
+    }: {
+      # hard coded fallbacks
+      channelName = "nixpkgs";
+      system = defaultSystem or "x86_64-linux";
```
Pacman99, line 36:
```suggestion
    }: {
      # hard coded fallbacks
      inherit channelName system;
```
How about do that instead, so you can still set the defaults up top, dropping the need for `null` and still allowing `defaultHostAttrs` to get priority over some things.

systemFlake.nix
Pacman99:
```suggestion
      modules = sharedModules ++ defaultHostAttrs.modules or [] ++ modules;
      extraArgs = sharedExtraArgs // defaultHostAttrs.extraArgs or {} // extraArgs;
```

systemFlake.nix
```diff
@@ -25,25 +25,29 @@
...
-    , system ? defaultSystem
     , extraArgs ? {}
-    , ...
-    }: defaultHostAttrs
-      // { 
-        inherit channelName system; 
-        modules = sharedModules ++ modules;
-        extraArgs = sharedExtraArgs // extraArgs;
-      };
+  } @ args: args;
```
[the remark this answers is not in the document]
blaggacao: Thanks for catching this!

systemFlake.nix
```diff
@@ -25,25 +25,29 @@
...
+
+  mergeHostAttrs = defaultAttrs: attrs: let
+    defaultTo = lhs: rhs: if rhs != null then rhs else lhs;
+  in
+  {
+    channelName = defaultTo (defaultTo "nixpkgs" defaultAttrs.channelName) attrs.channelName;
+    system = defaultTo (defaultTo defaultSystem defaultAttrs.system) attrs.system; # replace deaultSystem with x86_64-linux
+    modules = modules ++ defaultAttrs.modules ++ sharedModules;
+    extraArgs = extraArgs // defaultAttrs.extraArgs // sharedExtraArgs;
+  };
```
Pacman99, line 43: Your again implementing basic nix language features like `?` and `//` when its not necessary. Combining into one function makes things simpler.

systemFlake.nix
```diff
@@ -25,25 +25,29 @@
...
```
Pacman99, line 43:
```suggestion
evalHostArgs = 
  { channelName ? "nixpkgs"
  , system ? defaultSystem
  , modules ? [ ]
  , extraArgs ? {}
  }:  
  { 
    inherit channelName system; 
    modules = sharedModules ++ defaultHsotAttrs.modules ++ modules;
    extraArgs = sharedExtraArgs // defaultHostAttrs.extraArgs // extraArgs;
  }
```

Pacman99: obsoletes `validateHostArgs` since we can validate in same function.

systemFlake.nix
blaggacao: > Combining into one function makes things simpler.

That would necessarily mean mixing the preprocessing of what's _externally_ fed to the API and the actual business logic.

I can't conform such to the idea of contract validator being _separate_ from logic.

systemFlake.nix
blaggacao: I don't want external API contract enforcement be obsoleted, or made implicit, for that matter.

systemFlake.nix
```diff
@@ -24,18 +25,34 @@
 }@args:
 
 let
+  # clean dirty api input / ensure for a minimum that all expected attrs are present
+  validateHostAttrs = {
+      channelName ? null
+    , system ? null
+    , modules ? []
+    , extraArgs ? {}
+  } @ args: { inherit channelName system modules extraArgs; };
```
blaggacao, line 34: This accounts for the fact that we cannot do at the top:

```nix
{
  defaultHostAttrs ? { 
      channelName ? null
    , system ? null
    , modules ? []
    , extraArgs ? {}
  }
, ...
} :
```

It also brings each host's `profile` in a consistent shape.

blaggacao:
I'm abandoning this PR (which was meant as a fix) in favor of #22 which is meant as an implementation of its own right.

PR #22, opened 2021-04-08
add: defaultHostsAttrs (fixed)

blaggacao:
###### https://github.com/gytis-ivaskevicius/flake-utils-plus/commit/faa78550684ffbce1aab0f1df7d257f74f231588 unintentionally introduced non-functional business logic into `staging`. To keep staging clean, I reverted it in https://github.com/gytis-ivaskevicius/flake-utils-plus/commit/fe983e5a9cd4039b48e03f7c7e88ac750ebb2457 until we find a better solution. This is an attempt to improve on the previous implementation. As such, this supersedes #21, which was trying to be a fix only.

---

As the `fup` api grows, we need a better way to ensure consistency.

Currently top level has a mix of host-specific and channel-specific defaults. However, we expect host specific defaults to grow in the near future (eg. #17) and it would not be in order to just clutter the top level arguments.

This PR introduces a (reusable) API container and thereby transparently exposes the current `profile/host` `fup`-API-objet. For clarity, it currently is:
```nix
{
  profile = {
    channelName = "...";
    system = "...";
    extraArgs = {};
    modules = [];
  };
}
```

This PR generalizes this API, implements a minimal validation (expected args of the profile/host API object are present and unexpected args aren't) and consequentially, applies it analogously to a new `defaultHostAttrs`.

`defaultsHostAttrs` deprecated `defaultSystem`, because there is no way to prevent `defaultSystem` from becoming an obsolete no-op in the case that also `defaultSystem.system` is set. That would result in a very unclean API.

However, as `modules` and `extraArgs` obey merging semantics, `sharedModules` (`add`) & `sharedExtraArgs` (`merge`) are preserved as a convenient way of declaration and also to underline the _shared_ nature of them. And also for back-compat.

The exact merging semantics are encapsualted within `mergeHostAttrs` and are:
- system & channelName: more specific declartion's override
- modules: all declarations add
- extraArgs: declarations are merged and more specific declarations override in case of conflicting keys

`mergeHostAttrs` also implements the hard coded defaults for `channelName` and `system` (currently `defaultSystem` is only marked deprecated).

---

While working on this, I also realized that `profile` represents a `fup` specific api container, and does not represent a host config or a host profile. Hence, it might be in order to evaluate how we should properly name the following attribute set in a future discussion.
```nix
{
    channelName = "...";
    system = "...";
    extraArgs = {};
    modules = [];
}

### DIFF
--- a/systemFlake.nix
+++ b/systemFlake.nix
@@ -7,6 +7,7 @@
 
 , nixosConfigurations ? { }
 , sharedExtraArgs ? { }
+, defaultHostAttrs ? { }
 , nixosProfiles ? { } # will be deprecated soon, use nixosHosts, instead.
 , nixosHosts ? nixosProfiles
 , channels ? { }
@@ -24,18 +25,29 @@
 }@args:
 
 let
+  evalHostArgs =
+    { channelName ? "nixpkgs"
+    , modules ? []
+    , system ? defaultSystem
+    , extraArgs ? {}
+    , ...
+    }: defaultHostAttrs
+      // { 
+        inherit channelName system; 
+        modules = sharedModules ++ modules;
+        extraArgs = sharedExtraArgs // extraArgs;
+      };
+
   inherit (flake-utils-plus.lib) eachSystem;
 
   optionalAttrs = check: value: if check then value else { };
-
-  channelNameFromProfile = profile: profile.channelName or "nixpkgs";
-  systemFromProfile = profile: profile.system or defaultSystem;
 
   otherArguments = builtins.removeAttrs args [
     "defaultSystem"
     "sharedExtraArgs"
     "inputs"
     "nixosHosts"
+    "defaultHostAttrs"
     "channels"
     "channelsConfig"
     "self"
@@ -51,12 +63,12 @@ let
     "checksBuilder"
   ];
 
-  nixosConfigurationBuilder = hostname: profile: (
+  nixosConfigurationBuilder = hostname: profile: 
+    let hostAttrs = evalHostArgs profile; in
     # It would be nice to get `nixosSystem` reference from `selectedNixpkgs` but it is not possible at this moment
-    inputs."${channelNameFromProfile profile}".lib.nixosSystem (genericConfigurationBuilder hostname profile)
-  );
+    inputs."${hostAttrs.channelName}".lib.nixosSystem (genericConfigurationBuilder hostname hostAttrs);
 
-  getNixpkgs = profile: self.pkgs."${systemFromProfile profile}"."${channelNameFromProfile profile}";
+  getNixpkgs = profile: self.pkgs."${profile.system}"."${profile.channelName}";
 
   genericConfigurationBuilder = hostname: profile: (
     let selectedNixpkgs = getNixpkgs profile; in
@@ -87,9 +99,8 @@ let
           ];
         })
       ]
-      ++ sharedModules
-      ++ (profile.modules or [ ]);
-      extraArgs = { inherit inputs; } // sharedExtraArgs // profile.extraArgs or { };
+      ++ profile.modules;
+      extraArgs = { inherit inputs; } // profile.extraArgs;
     }
   );
 in
--- a/systemFlake.nix
+++ b/systemFlake.nix
@@ -75,15 +75,15 @@ let
     extraArgs = sharedExtraArgs // defaultAttrs.extraArgs // extraArgs;
   };
 
-  nixosConfigurationBuilder = hostname: profile: 
+  nixosConfigurationBuilder = hostname: hostAttrs: 
     let hostAttrs = mergeHostAttrs (validateHostAttrs defaultHostAttrs) (validateHostAttrs profile); in
     # It would be nice to get `nixosSystem` reference from `selectedNixpkgs` but it is not possible at this moment
     inputs."${hostAttrs.channelName}".lib.nixosSystem (genericConfigurationBuilder hostname hostAttrs);
 
-  getNixpkgs = profile: self.pkgs."${profile.system}"."${profile.channelName}";
+  getNixpkgs = hostAttrs: self.pkgs."${hostAttrs.system}"."${hostAttrs.channelName}";
 
-  genericConfigurationBuilder = hostname: profile: (
-    let selectedNixpkgs = getNixpkgs profile; in
+  genericConfigurationBuilder = hostname: hostAttrs: (
+    let selectedNixpkgs = getNixpkgs hostAttrs; in
     {
       inherit (selectedNixpkgs) system;
       modules = [
@@ -111,8 +111,8 @@ let
           ];
         })
       ]
-      ++ profile.modules;
-      extraArgs = { inherit inputs; } // profile.extraArgs;
+      ++ hostAttrs.modules;
+      extraArgs = { inherit inputs; } // hostAttrs.extraArgs;
     }
   );
 in
--- a/systemFlake.nix
+++ b/systemFlake.nix
@@ -25,20 +25,27 @@
 }@args:
 
 let
-  # clean dirty api input / ensure for that all expected, but no extra attrs are present
+  # ensure for that all expected, but no extra attrs are present
   validateHostAttrs = {
       channelName ? null
     , system ? null
     , modules ? []
     , extraArgs ? {}
-  } @ args: { inherit channelName system modules extraArgs; };
+  }: { inherit channelName system modules extraArgs; };
 
   mergeHostAttrs = defaultAttrs: attrs: let
-    defaultTo = lhs: rhs: if rhs != null then rhs else lhs;
+    # convoluted nix-kell, but clean(er) english-german below :-)
+    _ = x: op: y: op x y;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: systemFlake.nix [hunk 1]
```diff
@@ -24,18 +25,34 @@
...
+  mergeHostAttrs = defaultAttrs: attrs: let
+    defaultTo = lhs: rhs: if rhs != null then rhs else lhs;
+  in
+  {
+    channelName = defaultTo (defaultTo "nixpkgs" defaultAttrs.channelName) attrs.channelName;
+    system = defaultTo (defaultTo defaultSystem defaultAttrs.system) attrs.system; # replace deaultSystem with x86_64-linux
+    modules = modules ++ defaultAttrs.modules ++ sharedModules;
+    extraArgs = sharedExtraArgs // defaultAttrs.extraArgs // extraArgs;
+  };
+
```
Pacman99, line 45:
```suggestion
evalHostArgs = 
  { channelName ? "nixpkgs"
  , system ? defaultSystem
  , modules ? [ ]
  , extraArgs ? {}
  }:  
  { 
    inherit channelName system; 
    modules = sharedModules ++ defaultHsotAttrs.modules ++ modules;
    extraArgs = sharedExtraArgs // defaultHostAttrs.extraArgs // extraArgs;
  }
```
And `evalHostArgs` can still be called with `defaultHostAttrs // hostAttrs`, so // takes care of channelName and system and we can take care of modules/extraArgs manually, since those are a bit more nuanced.
To bring the discussion context here: I would prefer validation done in the outside of the function(arguments) and logic done on the inside. I'm not a fan of separating the two. @blaggacao wanted to separate the concerns to be clearer, because we was against doing logic like `channelName ? "nixpkgs"` in the same function that deals with merging. So his argument is validation and merging should not be done in the same function.

Overall goal here is to ensure contract validation for `defaultHostAttrs` and `hostAttrs`. I think both pieces of code accomplish that.

Pacman99 (changes requested):
Other than the implementation, I like this PR overall. And its good that you updated the examples appropriately, I didn't do that all too well.

systemFlake.nix
blaggacao: `defaultHostAttrs // hostAttrs` this not intended behaviour, actually.

I intentionally expect for:
```nix
{
  defaultHostAttrs.modules = [ a ];
  hostAttrs.modules = [ b ];
}
```
to become
```
{
  hostAttrs.module = [ a b ];
}
```

As far as I can understand the suggestion from merely reading it, we would arrive at:
```nix
{
  hostAttrs.module = [ b ];
}
```

That is not intended behavior of this PR. Am I missing something?

Pacman99: but then thats what these lines are for:
```nix
modules = sharedModules ++ defaultHsotAttrs.modules ++ modules;
extraArgs = sharedExtraArgs // defaultHostAttrs.extraArgs // extraArgs;
```
So the more intricate ones we merge manually. And importing modules twice isn't a concern, since module system deals with that.

Pacman99: So your right in that the `modules` that gets passed to `evalHostArgs` woould be `[ b ]` but then we do the merge again for modules, so it would return modules as `[ a b ]`.

blaggacao: Ah, now I think I spotted it. Is there a way to make this more obvious? (with a little less indirection and "double floors")

Pacman99: I'm not sure, I think this is the best way to do it. If you try and do it in the arguments, you end up with the problem that it gets overridden and not merged. I mean this is what you did in `mergeHostArgs` I'm just suggesting to do it in the same function as where defaults are set.

blaggacao: I mean is there just a way to _not_ tricking the reader into thinking that `rhs` will override `lhs` when doing `defaultHostAttrs // hostAttrs`? Like to accommodate a bit the average (non-expert) reader's attention span, that might just stop here and say: "aha, it's overridden then. Makes sense, it's called 'default', anyhow" (zoink!).

Pacman99: Ahh I think I finally now get your reasoning for `mergeHosts`. 
I think my suggestion is more accommodating to the reader that only looks at the top level attributes and `evalHostArgs`, its really clear whats going on, while your code requires a bit of hoops to jump through to see whats happening. 
Whereas for the reader at `defaultHostAttrs // hostAttrs` it might be confusing, but to be fair if you've read the top code I think it would be clear at this point. 
And I still think we should care more about the former reader than the latter, since most users and even contributors, likely won't go far enough into implementation to get confused by that one line.

blaggacao: I honestly did get confused. Now, I'm not sure what kind of read I am. :stuck_out_tongue_winking_eye: 

Let me see, what can be done.

blaggacao: @Pacman99

I think we can agree, that we want to avoid `defaultHostAttrs // hostAttrs` on the calling site, since it elicits a mislead interpretation of the code, a trap, we ourselves by the way have fallen into, previously.

Can you maybe adapt your suggestion accordingly to see if we can find a pareto-efficient solution to your concern?

blaggacao: So I addressed the awkward reading of `defaultTo`, which was plain bad English. I hope with this change, the code reads more easily, and the merge semantics are effortlessly evident to any class of reader. The reader is not expected to be forced to understand my "nix-kell" in-depth, but to take natural language which is construed with its help at face value.

systemFlake.nix
```diff
@@ -51,15 +58,32 @@ let
...
-  nixosConfigurationBuilder = hostname: profile: (
+  mergeHostAttrs = lhs: rhs: let
+    # convoluted nix-kell, but clean(er) english-german below :-)
+    _ = x: op: y: op x y;
+    oder = lhs: rhs: if lhs != null then lhs else rhs;
+
+    rhs.channelName' = _ rhs.channelName;
+    lhs.channelName' = _ lhs.channelName;
+    rhs.system' = _ rhs.system;
+    lhs.system' = _ lhs.system;
```
gytis-ivaskevicius, line 69: Now that's interesting :thinking:

systemFlake.nix
```diff
@@ -1,12 +1,13 @@
...
 
 { self
-, defaultSystem ? "x86_64-linux"
+, defaultSystem ? "x86_64-linux" # will be deprecated soon use defaultHostAttrs.system instead
 , supportedSystems ? flake-utils-plus.lib.defaultSystems
 , inputs
 
 , nixosConfigurations ? { }
 , sharedExtraArgs ? { }
+, defaultHostAttrs ? { }
```
gytis-ivaskevicius, line 10: Not a fan of the name, how do these two sound to you?
`hostDefaults`
`defaultHostConfig`

gytis-ivaskevicius (approved):

Pacman99 (commented):
I just can't approve this when validateHostArgs and mergeHostArgs will always be called together, and combining the functions would reduce code length and complexity.

If you all think its ok then thats fine, I'm just against the current implementation.

systemFlake.nix
blaggacao: I've seen such language features in haskell and I was frustrated that I could not do:
`a or b or c`.

systemFlake.nix
Pacman99: I like both, I vote for `hostDefaults`

systemFlake.nix
Pacman99: actually nix has operator overloading! Not sure if it helps here, but you can do things llike `__lessthan = `

systemFlake.nix
Pacman99: But both names don't communicate how the modules under `defaultHostAttrs` is shared. default kind of sounds like it gets overridden. The examples are good though, so I don't think it will be much of an issue.

systemFlake.nix
gytis-ivaskevicius: > I've seen such language features in haskell and I was frustrated that I could not do: `a or b or c`.

Yeah, I realize what you were trying to do here, quite smart tho now that I look at it - at very least I would not mind better naming or comment or two

> actually nix has operator overloading! Not sure if it helps here, but you can do things like __lessthan =
That's cool, I did not know that!

Pacman99: Yeah this code is quite smart for what its trying to accomplish

systemFlake.nix
blaggacao:
```
nix-repl> _ = x: op: y: op x y

nix-repl> __or = lhs: rhs: if lhs != null then lhs else rhs

nix-repl> or 1 2
error: syntax error, unexpected OR_KW
```

I conclude `or` is not an operator and for keywords it doesn't work.

blaggacao: Done: renamed to `hostDefaults` & `hosts` accordingly

systemFlake.nix
blaggacao: I've tried al possible combination. It looks as if overloading `__or` does not as intended.

blaggacao:
After a few fixups, I'm going to merge this, but I'd like to get #24 going, so I can actually validate everything. Thus far I go seems promising, though.

PR #23, opened 2021-04-09
ref: `profile` -> `hostAttrs`

blaggacao:
Rationale:

For internal coding purposes only, the following attribute set does
not require an abstract name (such as `profile`). Moreover so, as
"profile" has a distinct legacy meaning in the nix ecosystem.
```nix
 {
    channelName = "...";
    system = "...";
    extraArgs = {};
    modules = [];
 }
```
Instead, this PR renames it to a banal and descriptive `hostAttrs`. This
naming does also better align with the top level `defaultHostAttrs`.

### DIFF
--- a/systemFlake.nix
+++ b/systemFlake.nix
@@ -75,15 +75,15 @@ let
     extraArgs = sharedExtraArgs // defaultAttrs.extraArgs // extraArgs;
   };
 
-  nixosConfigurationBuilder = hostname: profile: 
+  nixosConfigurationBuilder = hostname: hostAttrs: 
     let hostAttrs = mergeHostAttrs (validateHostAttrs defaultHostAttrs) (validateHostAttrs profile); in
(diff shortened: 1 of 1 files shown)


Pacman99:
Can we include `nixosHosts` -> `hosts` in this PR, so it doesn't have to be a part of #17. And actually I'd like to propose the name `configs` since that covers home-manager too, and the flake output is called `configurations` for all systems, so I think that works the best.

blaggacao:
I agree with everything you said.

Since this is a very targeted PR and only pretends to solve the stated issue, I'd rather leave it as it is.

Since the code still contains `nixosConfigurationBuilder` and in general `nixos*` namespacing, I wouldn't regard renaming `nixosHosts` to `hosts` as an atomic change without also solving the prefix over there.

gytis-ivaskevicius (approved):

PR #26, opened 2021-04-10
imp: add specialArgs

blaggacao:
anything that is going to be passed to an `imports = []` necesarily
needs to be passed as specialArgs. This ensures it does not become
part of the module system (extraArgs = _module.args) itself.

Since it is not part of the module system, modules can evaluate based on
inputs, otherwise no fixed point would exists (infinit recursion).

This is a way to encapsulate profiles into bundles that are accessible
like so:

inputs = [ bundles.home ];

### DIFF
--- a/systemFlake.nix
+++ b/systemFlake.nix
@@ -36,7 +36,8 @@ let
     , builder ? null
     , modules ? [ ]
     , extraArgs ? { }
-    }: { inherit channelName system output builder modules extraArgs; };
+    , specialArgs ? { }
+    }: { inherit channelName system output builder modules extraArgs specialArgs; };
 
   mergeHosts = lhs: rhs:
     let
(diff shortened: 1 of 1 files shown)


systemFlake.nix
```diff
@@ -36,7 +36,8 @@ let
     , builder ? null
     , modules ? [ ]
     , extraArgs ? { }
-    }: { inherit channelName system output builder modules extraArgs; };
+    , specialArgs ? { }
+    }: { inherit channelName system output builder modules specialArgs; };
```
Pacman99, line 40:
```suggestion
    }: { inherit channelName system output builder modules extraArgs specialArgs; };
```

Pacman99 (changes requested):
Outside of this it looks good

Pacman99 (approved):
I think this should be exposed, you can use it to set `modulesPath` or send groups of modules to a host, ex: mobile-nixos's flake needs to be used like this.

gytis-ivaskevicius (approved):
looks good to me :)

PR #28, opened 2021-04-12
implement builtinPackagesBuilder

### LINKED ISSUE #12: sharing model
https://github.com/gytis-ivaskevicius/flake-utils-plus/blob/51cb739c9c9c2258bc70747eb7bc22975ae244bd/systemFlake.nix#L5

What is the rational of defaulting to `inputs` rather than have people make use of inputs in their top level flake?

The `inputs` thought model as `extraArgs` neglects the "receptionist" thought model of the flake.nix.
I think `flake.nix` as receptionist is an important feature of the distributed future of flakes, "to show new guests around": Hans visit's James' repo on github. flake.nix gives him the map and tells at least the outline of the story.

What cases cannot be covered by receiving inputs in flake.nix and dispatching eg overlays via `input.otherflake.overlay` or `inputs.otherflake.nixosModules`?

I think it was not an afterthought to make those two public static api of flakes themselves, in the first place by the designers of flakes.

One thing that comes to my mind ...

In a sharing model based on `overlay` it is not always trivial to pull in the following:

```nix
let
   bob = {
     kakoune = {};
     chrome = {};
   };
   alice = {
     kakoune = {};
     chrome = {};
   };
in
[
  bob.kakoune
  alice.chrome
]
```

`flake-utils.lib.simpleFlake` solves this by [implicitly](https://github.com/numtide/flake-utils/blob/b2c27d1a81b0dc266270fa8aeecebbd1807fc610/examples/simple-flake/overlay.nix#L4) doing a [namespacing of overlays](https://github.com/numtide/flake-utils/blob/b2c27d1a81b0dc266270fa8aeecebbd1807fc610/simpleFlake.nix#L52).

Should we really go down that route and just globally pass through inputs, by default?
&rarr; What this does _not_ do is expose and promote encapsulation via the top level flake api.

/cc @Pacman99 @nrdxp

blaggacao:
##### closes #12 

This is a convenient function that enables sane default sharing of channel overlays:

- It filters on packages per channel, that are also defined in overlays
- therefore, only "custom" stuff that adds potential value to the world is exported

```nix
{
  self.packages."<system>" = {
    # only packages per system who's checks do not trivially fail:
    # - system compatible
    # - a derivation 
    # - meta.broken == false
    # and only packages that are defined by:
    # - some overlay 
    # - on some channel.
    # that includes classical override variants of packages, 
    # as well as new package definitions.
    kakoune = { ... };
  };
}
```

----

checpoint w/o the following:

- cleanup / refactoring
- docstrings
- tackle nixpkgs.lib-is-not-cheap-to-import-issue
- discuss packageBuilder api extension with `channels`

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -4,19 +4,6 @@
   inputs.flake-utils.url = "github:numtide/flake-utils";
 
   outputs = { self, flake-utils }:
-    let
-      removeSuffix = suffix: str:
-        let
-          sufLen = builtins.stringLength suffix;
-          sLen = builtins.stringLength str;
-        in
-        if sufLen <= sLen && suffix == builtins.substring (sLen - sufLen) sufLen str then
-          builtins.substring 0 (sLen - sufLen) str
-        else
-          str;
-
-      genAttrs' = func: values: builtins.listToAttrs (map func values);
-    in
     rec {
 
       nixosModules.saneFlakeDefaults = import ./modules/saneFlakeDefaults.nix;
@@ -26,12 +13,9 @@
         repl = ./repl.nix;
         systemFlake = import ./systemFlake.nix { flake-utils-plus = self; };
 
-        modulesFromList = paths: genAttrs'
-          (path: {
-            name = removeSuffix ".nix" (baseNameOf path);
-            value = import path;
-          })
-          paths;
+        exporter = {
+          modulesFromListExporter = import ./modulesFromListExporter.nix { flake-utils-plus = self; };
+        };
 
         patchChannel = system: channel: patches:
           if patches == [ ] then channel else
--- a/flake.nix
+++ b/flake.nix
@@ -14,6 +14,7 @@
         systemFlake = import ./systemFlake.nix { flake-utils-plus = self; };
 
         exporter = {
+          overlaysFromChannelsExporter = import ./overlaysFromChannelsExporter.nix { flake-utils-plus = self; };
           modulesFromListExporter = import ./modulesFromListExporter.nix { flake-utils-plus = self; };
         };
 
--- a/flake.nix
+++ b/flake.nix
@@ -13,6 +13,10 @@
         repl = ./repl.nix;
         systemFlake = import ./systemFlake.nix { flake-utils-plus = self; };
 
+        builder = {
+          packagesFromOverlaysBuilderConstructor = import ./packagesFromOverlaysBuilderConstructor.nix { flake-utils-plus = self; };
+        };
+
         exporter = {
           overlaysFromChannelsExporter = import ./overlaysFromChannelsExporter.nix { flake-utils-plus = self; };
           modulesFromListExporter = import ./modulesFromListExporter.nix { flake-utils-plus = self; };
--- a/moduleFromListExporter.nix
+++ b/moduleFromListExporter.nix
@@ -1,0 +1,29 @@
+{ flake-utils-plus }: let
+
+modulesFromListExporter =
+
+  let
+
+    removeSuffix = suffix: str:
+      let
+        sufLen = builtins.stringLength suffix;
+        sLen = builtins.stringLength str;
+      in
+      if sufLen <= sLen && suffix == builtins.substring (sLen - sufLen) sufLen str then
+        builtins.substring 0 (sLen - sufLen) str
+      else
+        str;
+
+    genAttrs' = func: values: builtins.listToAttrs (map func values);
+
+  in
+
+    paths: genAttrs'
+      (path: {
+        name = removeSuffix ".nix" (baseNameOf path);
+        value = import path;
+      })
+      paths;
+
+in
+modulesFromListExporter
--- a/moduleFromListExporter.nix
+++ b/moduleFromListExporter.nix
@@ -1,24 +1,25 @@
-{ flake-utils-plus }: let
+{ flake-utils-plus }:
+let
 
   modulesFromListExporter = paths:
     /**
-    Synopsis: modulesFromListExporter _paths_
+      Synopsis: modulesFromListExporter _paths_
 
-    paths:    [ <path> <path> ]
+      paths:    [ <path> <path> ]
 
-    Returns an attribute set of modules from a list of paths by converting
-    the path's basename into the attribute key.
+      Returns an attribute set of modules from a list of paths by converting
+      the path's basename into the attribute key.
 
-    Example:
+      Example:
 
-    paths:    [ ./path/to/moduleA.nix ./path/to/moduleBfolder ]
+      paths:    [ ./path/to/moduleA.nix ./path/to/moduleBfolder ]
 
-    {
+      {
       moduleA = import ./path/to/moduleA.nix;
       moduleBfolder = import ./path/to/moduleBfolder;
-    }
+      }
 
-    **/
+      **/
 
     let
 
@@ -36,12 +37,12 @@
 
     in
 
-      genAttrs'
-        (path: {
-          name = removeSuffix ".nix" (baseNameOf path);
-          value = import path;
-        })
-        paths;
+    genAttrs'
+      (path: {
+        name = removeSuffix ".nix" (baseNameOf path);
+        value = import path;
+      })
+      paths;
 
 in
 modulesFromListExporter
--- a/moduleFromListExporter.nix
+++ b/moduleFromListExporter.nix
@@ -1,29 +1,47 @@
 { flake-utils-plus }: let
 
-modulesFromListExporter =
+  modulesFromListExporter = paths:
+    /**
+    Synopsis: modulesFromListExporter _paths_
 
-  let
+    paths:    [ <path> <path> ]
 
-    removeSuffix = suffix: str:
-      let
-        sufLen = builtins.stringLength suffix;
-        sLen = builtins.stringLength str;
-      in
-      if sufLen <= sLen && suffix == builtins.substring (sLen - sufLen) sufLen str then
-        builtins.substring 0 (sLen - sufLen) str
-      else
-        str;
+    Returns an attribute set of modules from a list of paths by converting
+    the path's basename into the attribute key.
 
-    genAttrs' = func: values: builtins.listToAttrs (map func values);
+    Example:
 
-  in
+    paths:    [ ./path/to/moduleA.nix ./path/to/moduleBfolder ]
 
-    paths: genAttrs'
-      (path: {
-        name = removeSuffix ".nix" (baseNameOf path);
-        value = import path;
-      })
-      paths;
+    {
+      moduleA = import ./path/to/moduleA.nix;
+      moduleBfolder = import ./path/to/moduleBfolder;
+    }
+
+    **/
+
+    let
+
+      removeSuffix = suffix: str:
+        let
+          sufLen = builtins.stringLength suffix;
+          sLen = builtins.stringLength str;
+        in
+        if sufLen <= sLen && suffix == builtins.substring (sLen - sufLen) sufLen str then
+          builtins.substring 0 (sLen - sufLen) str
+        else
+          str;
+
+      genAttrs' = func: values: builtins.listToAttrs (map func values);
+
+    in
+
+      genAttrs'
+        (path: {
+          name = removeSuffix ".nix" (baseNameOf path);
+          value = import path;
+        })
+        paths;
 
 in
 modulesFromListExporter
--- a/overlaysFromChannelsExporter.nix
+++ b/overlaysFromChannelsExporter.nix
@@ -1,0 +1,61 @@
+{ flake-utils-plus }: let
+
+ overlaysFromChannelsExporter = channels:
+  /**
+  Synopsis: overlaysFromChannelsExporter _channels_
+
+  channels: channels.<name>.overlays
+
+  Returns an attribute set of all packages defined in an overlay by any channel
+  intended to be passed to be exported via _self.overlays_. This method of
+  sharing has the advantage over _self.packages_, that the user will instantiate
+  overlays with his proper nixpkgs version, and thereby significantly reduce their system's
+  closure as they avoid depending on entirely different nixpkgs versions dependency
+  trees. On the flip side, any caching that is set up for one's packages will essentially
+  be useless to users.
+
+  It can happen that an overlay is not compatible with the version of nixpkgs a user tries
+  to instantiate it. In order to provide users with a visual clue for which nixpkgs version
+  an overlay was originally created, we prefix the channle name: "<channelname>/<packagekey>".
+  In the case of the unstable channel, this information is still of varying usefulness,
+  as effective cut dates can vary heavily between repositories.
+
+  Example:
+
+  overlays = [
+    "unstable/development" = final: prev: { };
+    "nixos2009/chromium" = final: prev: { };
+    "nixos2009/pythonPackages" = final: prev: { };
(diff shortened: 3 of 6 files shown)


gytis-ivaskevicius:
I won't pretend that I understand what you are doing here - May I see some actual examples/use cases? 

Currently, overlays are applied to channels and packages/system configurations/repl are being populated from there - and I do quite like this approach, data flow is nice and simple.

As for exposing overlays - it is done by standard overlay/overlays attributes, just like in normal flakes. https://github.com/gytis-ivaskevicius/nixfiles/blob/master/flake.nix#L148

blaggacao:
> May I see some actual examples/use cases?

I definitly owe better context, still. I'll give a peak, here:

The problem this PR tries to solve goes by this storyline:

What is useful content of `self.packages`? `self.packages` is a way to share one's art with others so for example, if I'd packages latest FRRouting, which is not available on Nixpkgs, then people might have an interest in using that.

One mediumly sane assumption _could_ be that everything for which I built an overlay might also be worth sharing with the world.

This PR builds on that assumption and provides a mean to export all packages that where declared through any of the channel overlays.

I still need to so the other half of the work on this like cleanup and some refactoring. If you are wary to accept such propoesl, I can very well understand you and no worries, I'd include it in devos, instead.

gytis-ivaskevicius:
Alright, I hear you, but let me raise this point:
Is `self.packages` actually a good way of sharing packages? Nix provides 2 types of packages of sorts - overlays and actual packages, the difference is that one depends on the maintenance of the existing repository. 

Sharing using overlays may not be entirely straight forward but in better solution due to these reasons:
1. Dependency (inputs) reusability
2. Dependency (inputs) updates

The downside is that it involves a few extra steps:
1. Apply overlay to your main channel
2. Install applied package (which name you might need to manually lookup)
3. If it works - all good, if it does not - apply the overlay to an unstable channel and install it from there

Upsides of `self.packages`:
1. One can do stuff like `nix run github:gytis-ivaskevicius/nixfiles#g-neovim`


I strongly dislike the concept of packages, and honestly, it makes me wonder if it would be possible for nix to have only overlays instead with the addition of "strong" and "weak" dependency concept where weak dependencies would be replaced depending on the user-specified channel and strong dependencies would be explicitly specified in the package maintainer flake 


Anyways, I don't mind this feature, but implementation and especially the user-facing side must be pretty sleek

blaggacao:
Ha, you have put my nacent thought in even better words, here.

For the "strong" & "weak" dependency idea:

In order to give consumer at least a visual.clue about a what channel aproximately an overlay was made for, we planned to export overlays primarily as:

```nix
{
  self.overlays = {
    "unstable/kakoune" = final: prev: { };
    "nixos-20.09/firefox" = final: prev: { };
  };
}
```

gytis-ivaskevicius:
I had in mind a much more fine graned idea of weak/strong dependencies, way too complex and out of scope for fup and devos. But yeah.. It does not really matter

Also, I doubt that overlay exporting like that is necessary especially since there is a big difference between unstable 1 day ago vs unstable 6months ago

blaggacao:
Maybe you are right and we should discourage the use of `self.packages` in the context of a nixos sharing model.

As I wrote my peak, this possibiloty was already dawning in my head.

It is still a good tool for software projects, but it is arguably less useful for the nixos sharing model.

blaggacao:
You are also right about unstabld, a better semantic might be, if we can source that distincion meaningfully from the inputs...

```nix
​{
  ​self​.​overlays​ ​=​ {
    ​"​kakoune​"​ ​=​ ​final​: ​prev​: { };
    ​"​nixos-20.09/firefox​"​ ​=​ ​final​: ​prev​: { };
  };
}
```

Pacman99:
There is one big advantage of packages, and I don't think any of these messages touched it. Packages are entirely self contained. Overlays depend on a specific version of nixpkgs, because they build the package using the version of nixpkgs given to them(_prev_: _final_:). This point was raised already.

Whereas when I get a package from a flake, it has already been created with some version of nixpkgs. And while this might result in dependency reuse, I am guaranteed that building that package will work, because of nix. So in my opinion overlays are more convenient, but packages are a much better way of sharing your packages because they provide safety riding on nix's guarantees.

This is sort of like if I had tested firefox on nixpkgs/someCommit and told you that it worked. If you install firefox from that same nixpkgs commit, it WILL work for you. In the same vein I am creating my package with a version of nixpkgs that I know works, so therefore I can share it with much more safety. This is not true for overlays, because you are using an "untested" version of nixpkgs to build my package. And even if you prefer using overlays to share packages, we should still consider the advantages of sharing via packages.

systemFlake.nix
```diff
@@ -1,6 +1,66 @@
-{ flake-utils-plus }:
+{ flake-utils-plus }: let
 
-{ self
+ builtinPackagesBuilder = pkgs: channels:
```
Pacman99, line 3: In my opinion this function is complex/large enough to deserve its own file. But more importantly I think you should export it in flake-utils. It has a lot of value for others. I'm indifferent on whether this should be done by default. 

But I think this would be better done as a function exported by fup. And then its agreed to do it by default, then you can do `packagesBuilder ? flake-utils-plus.builtinPackagesBuilder`.

Although in this case perhaps it should be named differently. overlaysPackagesBuilder? autoPackagesBuilder?

blaggacao:
I suspect, a good argument to consider for the sharing model, though are distributed binary caches.

They would very likely only work on unmodified `self.packages` and wouldn't probably work on overlays.

Sure this goes at the expens of tons of GB in data, but in some low-optimized environments, this might simply afford us the better user experiemce.

(Just thinking loud)

systemFlake.nix
blaggacao: Good points. This is only my first iteration, and I definitly have half the worl still before me. :wink:

gytis-ivaskevicius: If it is getting moved to a separate file - you probably will need stuff like `optionals` and `mkIf` - probably a utils file should be created while we are at it. 

I'm fine with exposing publically via `self.lib.internal.<function-name>` or having that reference privately, it up to you guys to decide

blaggacao: > If it is getting moved to a separate file - you probably will need stuff like optionals and mkIf - probably a utils file should be created while we are at it.

I actually think it's a nice coding style of upstream flake-utils that each file can be entirely read by its own, without external references. While it _might_ duplicate a little code, it's much more polite to first time readers of the fup repo, and transcending this, it's probably good marketing for this repo.

systemFlake.nix
```diff
@@ -161,7 +160,7 @@ otherArguments
     optional = check: optionalAttrs (check != null);
   in
   { inherit pkgs; }
-  // optional packagesBuilder { packages = packagesBuilder pkgs; }
+  // optional packagesBuilder { packages = packagesBuilder pkgs channels; }
```
Pacman99, line 163: Why does channels need to be passed here? Is there any way this can be avoided, perhaps by passing channels as the first argument to overlaysPackagesBuilder, and users can do the same - pass their channels to create the packagesBuilder.
I don't like having to change the api for this builder.

overlaysPackagesBuilder.nix
```diff
@@ -0,0 +1,106 @@
...
+
+  Finally, only passes packages through the seive that are prefixed with a top level
+  key exposed by any of the overlays. Since overlays override (and do not emrge)
+  top level attributes, by filtering on the prefix, an overlay's entire packages
+  tree will be correctly captured.
+
+  Example:
+
+  pkgs'.<system> = {
+    "development/kakoune" = { ... };
```
Pacman99, line 30: To be clear this means that these packages were put into pkgs as `pkgs.development.kakoune` right?

Pacman99 (approved):
I really like this! It has a lot of potential to improve user's sharing model.

systemFlake.nix
blaggacao: I also don't like it. I'm going to think a little harder. Thanks!

overlaysPackagesBuilder.nix
blaggacao: Correct, that is: the overlay key was named (for whatever reason) "development" and had an attribute set marked with `{ recurseForDerivations = true; }`

Typically that would have to look like so:

```nix
{
final: prev: {
  development = prev.recurseIntoAttrs {
    kakoune = {};
  };
}
}
```

If not setup this way, flattenTree would not recurse into the attribute set (just as any other standard nixpkgs tool in similar situations).

flake.nix
```diff
@@ -26,12 +13,14 @@
...
         systemFlake = import ./systemFlake.nix { flake-utils-plus = self; };
 
-        modulesFromList = paths: genAttrs'
-          (path: {
-            name = removeSuffix ".nix" (baseNameOf path);
-            value = import path;
-          })
-          paths;
+        builders = {
+          overlaysPackagesBuilder = import ./overlaysPackagesBuilder.nix { flake-utils-plus = self; };
```
blaggacao, line 17: `packagesFromOverlaysBuilderConstructor`

`Constructor` as it "constructs" a "builder" by injecting the `channels` dependency.

flake.nix
```diff
@@ -26,12 +13,14 @@
...
+        };
+
+        exporters = {
+          overlaysExporter = import ./overlaysExporter.nix { flake-utils-plus = self; };
```
blaggacao, line 21: `overlaysFromChannelsExporter`

blaggacao (commented):
Notes to self

overlaysPackagesBuilder.nix
Pacman99: ok perfect thanks for the explanation. That makes a lot of sense.

flake.nix
Pacman99: For any name you do choose, I think `packagesBuilder` should be inside it just to be clear where this function is used and what its for.

Pacman99: I think the name is actually good, maybe just reorder fromOverlaysPackagesBuilder

blaggacao: It's analogous to the modulesFromListExporter / overlaysFromChannelExporter.

`modulesFromList` set the precedence here.

moduleFromListExporter.nix
```diff
@@ -0,0 +1,47 @@
...
+    Returns an attribute set of modules from a list of paths by converting
+    the path's basename into the attribute key.
+
+    Example:
+
+    paths:    [ ./path/to/moduleA.nix ./path/to/moduleBfolder ]
+
+    {
+      moduleA = import ./path/to/moduleA.nix;
+      moduleBfolder = import ./path/to/moduleBfolder;
```
blaggacao, line 18: While working on the docs for moduleFromListExporter, I was wondering if we would take good advice here and overload those with some metadata concerning either or a combination of:
- location
- ownership (by this flake's current world-wide unique commit hash)

Prospectively, those enable us to distinguish in a global namespace which modules to auto-export by default without the user in need to tell us.

@Pacman99 How could such metadata enrichment potentially look like at this position?

I'd very much favor to record ownership by a specific commit hash, here, since it seems universally useful and directly testable by a downstream user's flake's `self.rev`.

blaggacao:
```nix
{
  modulesFromListExporter = paths: owningRev:
}
```
(if that makes sense)

---

maybe also
```nix
{
  overlaysFromChannelsExporter = channels: owningRev:
}
```

Pacman99: You could use __functor for this. The overlay/module can go in __functor and metadata can be part of the attrset. I would say save this for future work however.

blaggacao:
@gytis-ivaskevicius did you already make your mind up on the general direction that this proposal is taking? In a sort is perpetuates the direction the `modulesFromList` previously has set and I believe it is a plausible (intermeadiate) answer to the sharing model we could encourage.

Intermediate, since we might well need to iterate on the details after collecting some more experience or as flakes themselves evolve.

overlaysFromChannelsExporter.nix
```diff
@@ -0,0 +1,61 @@
...
+    inherit (builtins) mapAttrs attrNames concatMap listToAttrs;
+    nameValuePair = name: value: { inherit name value; };
+
+    pathStr = builtins.concatStringsSep "/" path;
+
+    channelNames = attrNames channels;
+    overlayNames = overlay: attrNames ( overlay null null );
+
+    extractAndNamespaceEachOverlay = channelName: overlay:
+      map (ovrlName:
```
gytis-ivaskevicius, line 42: Can we have a better argument name than `ovrlName` :D

packagesFromOverlaysBuilderConstructor.nix
```diff
@@ -0,0 +1,111 @@
...
+    Example:
+
+    pkgs'.<system> = {
+      "development/kakoune" = { ... };
+      "development/vim" = { ... };
+    };
+
+    overlays' = [
+      "development"
+    ];
```
gytis-ivaskevicius, line 38: It would be cool to make this typesafe, something along the lines of:
```
    overlays' = [
      xyz.development
    ];
```
Just some food for thought, dont mind me :)

systemFlake.nix
```diff
@@ -16,7 +16,7 @@
 , sharedExtraArgs ? { }
 , sharedModules ? [ ]
 
-, packagesBuilder ? null
+, packagesBuilder ? (flake-utils-plus.lib.builder.packagesFromOverlayBuilderConstructor channels)
```
gytis-ivaskevicius, line 19: Not sure if this should be the default - might confuse some users

gytis-ivaskevicius (approved):
I would be very pleased with having all this as part of fup :+1: 

Few general comments:
- inconsistent use of `inherit (builtins) mapAttrs;` and `builtins.mapAttrs`
- very odd spacing
- some let/in statements and function calls don't make much sense - let/in syntax should be removed in a couple of locations

Also, another random idea: something along the lines of `$ nix run build development` could be implemented so users would have easier time building their aplications using Github actions

systemFlake.nix
blaggacao: Not sure either, will remove it for now. Being conservative might appeal to our audience.

packagesFromOverlaysBuilderConstructor.nix
blaggacao: It is in fact stringly typed on purpose for the (string) prefix matching to work after pkgs' tree has been flattened. I also wouldnt be able to tell what `xyz` could be, since overlays  define top level attributes on the `pkgs` namespace.

blaggacao:
I think I mostly addressed your concerns. I think `nix run build development` is +- enabled through `packagesFromOverlaysBuilderConstructor`, It's just that each package has to be called individually. I might assume your idea is to have the convenience of just having one single "build job". It would be actually interesting to explore how that could best work with the `jobs` or `hydraJobs` outputs (`job` for the sake of being agnostic).

gytis-ivaskevicius:
This code does not even parse, what's up with that? How does one even work on a feature when that is the case?

blaggacao:
I got tossed away from keyboard last night. I should have put that back into draft, strictly speaking. :wink:

blaggacao:
I'm going to merge this, since it's the last piece missing for us to rebase `devos` for which we have a +- plan to finalize it by this weekend. It also does not really hurt any existing workflow since it mostly adds features, and I don't know how to test it really properly (other than rebasing devos which will make intricate use of it).

To address the latter I opened #35, which should help with that, going forward.

PR #29, opened 2021-04-13
imp: ease nixpkgs dependency ban by using divnix/nixpkgs.lib light fork

blaggacao:
This enables us to use latest nixpkgs library functions without the "penalty" and confusion of accessing nixpkgs propper.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -2,8 +2,9 @@
   description = "Pure Nix flake utility functions";
 
   inputs.flake-utils.url = "github:numtide/flake-utils";
+  inputs.lib.url = "https://github.com/divnix/nixpkgs.lib";
 
-  outputs = { self, flake-utils }:
+  outputs = { self, flake-utils, lib }:
     let
       removeSuffix = suffix: str:
         let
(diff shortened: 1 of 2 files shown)


gytis-ivaskevicius (changes requested):
I can't approve additional dependency, especially one like this

blaggacao:
Completely get your point, I set up that library in general for projects that try to avoid importing nixpkgs, but suffer from so many library functions not yet available as builtins.

It's meant to alleviate a bit the burdon of reimplementing library functions across the aisle.

See: https://discourse.nixos.org/t/cheaper-nixpkgs-lib-for-flake-users/12444?u=blaggacao


No issues with barring this from entering `fup`.

Pacman99:
I think one other thing to consider about flake-utils(-plus) is that the lack of nixpkgs dependency is also somewhat of a principle. I think its cool that in the entire flake ecosystem, there will be these two - maybe more - repositories that never depend on something else. And while it doesn't seem like such a big deal right now, I think as flakes get bigger it could become more important. I'm not sure we have to worry about cyclic dependencies with flakes. But either way its just good practice for software ecosystems to have projects that act as a base/pillar upon which anyone can work on without having to think about what that project is dependent on.

PR #32, opened 2021-04-14
Cleanup + a couple of little bugs

gytis-ivaskevicius:


examples/fully-featured/myNixpkgsPatch.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/pkgs/top-level/all-packages.nix b/pkgs/top-level/all-packages.nix
+index 6dd393492c8..c8456afc28d 100644
+--- a/pkgs/top-level/all-packages.nix
++++ b/pkgs/top-level/all-packages.nix
+@@ -30930,4 +30930,6 @@ in
+   lc3tools = callPackage ../development/tools/lc3tools {};
+
+   zktree = callPackage ../applications/misc/zktree {};
++
++  flake-utils-plus-test = callPackage ../applications/misc/hello { };
```
gytis-ivaskevicius, line 10: Actually a proper patch to test patching functionality

repl.nix
```diff
@@ -6,6 +6,6 @@ in
 { inherit flake; }
 // flake
 // builtins
-// nixpkgs
+// flake.nixosConfigurations.${hostname}
 // nixpkgs.lib
-  // flake.nixosConfigurations.${hostname}
+  // nixpkgs
```
gytis-ivaskevicius, line 11: There was an issue that `flake.nixosConfigurations.<hostname>.pkgs` were overwriting `nixpkgs` which was not intended

systemFlake.nix
```diff
@@ -9,7 +9,6 @@
 , channelsConfig ? { }
 , sharedOverlays ? [ ]
 
-, nixosConfigurations ? { } # deprecate soon, no longer used or works
```
gytis-ivaskevicius, line 12: It was unused

systemFlake.nix
```diff
@@ -179,6 +168,6 @@ otherArguments
   // optional checksBuilder { checks = checksBuilder pkgs; }
```
gytis-ivaskevicius, line 168: There is an issue of builders not getting merged with actual arguments :/ I'd love to see such use-case working:
```patch
diff --git a/examples/fully-featured/flake.nix b/examples/fully-featured/flake.nix
index 5eb7ade..13071f9 100644
--- a/examples/fully-featured/flake.nix
+++ b/examples/fully-featured/flake.nix
@@ -139,6 +139,7 @@

       # All other values gets passed down to the flake
       overlay = import ./overlays;
+      packages.x86_64-linux.merge-with-packagesBuilder-test = self.pkgs.x86_64-linux.nixpkgs.hello;
       abc = 132;
       # etc
 ```
 Maybe you guys could think of some clean merge function for this? My solution would be quite lenghty

gytis-ivaskevicius: Dumb and simple idea came to my mind:
 `mapAttrs (system: _: packages1.${system} or {} // packages2.${system} or {} ) (packages1 // packages2)` 

looks good?

fufexan: There may be the same issue with `appsBuilder` and `apps`. Should probably fix that as well.

gytis-ivaskevicius: Yes, there are. Everything that depends on `<system>` is implemented in the same manner, but also fix is the same for all of these properties as well

PR #33, opened 2021-04-15
Improve merging of host arguments and flake outputs

Pacman99:
Create a merge function that can deal with attrsets and lists and use it for:
 - merging host arguments
 - configuration outputs - use it in foldHosts and allow for nixosConfigurations escape hatch

Also fix the issue of packagesBuilder overriding `otherArguments.packages`, by merging each output manually with mkOutput function. Also cleans up the output merging code, its just two strings passed to a function. This doesn't require the merge function.

### DIFF
--- a/examples/fully-featured/flake.nix
+++ b/examples/fully-featured/flake.nix
@@ -36,26 +36,21 @@
         system = "aarch64-linux";
         # Default channel to be used for `hosts` defaults to "nixpkgs"
         channelName = "unstable";
-        # Extra arguments to be passed to modules. Merged with sharedExtraArgs on its left hand side and the host's extraArgs on its right hand side
-        extraArgs = { foo = "foo"; };
-        # Default modules to be passed to all hosts. Equivalent to sharedModules (additive merge).
-        modules = [ ];
+        # Extra arguments to be passed to modules. Merged with host's extraArgs
+        extraArgs = { inherit utils inputs; foo = "foo"; };
+        # Default modules to be passed to all hosts.
+        modules = [
+          home-manager.nixosModules.home-manager
+          # Sets sane `nix.*` defaults. Please refer to implementation/readme for more details.
+          utils.nixosModules.saneFlakeDefaults
+          (import ./modules)
+          {
+            home-manager.useGlobalPkgs = true;
+            home-manager.useUserPackages = true;
+          }
+        ];
+
       };
-
-      # Extra arguments to be passed to modules. Defaults to `{ inherit inputs; }`
-      sharedExtraArgs = { inherit utils inputs; };
-
-      # Shared modules/configurations between `hosts`
-      sharedModules = [
-        home-manager.nixosModules.home-manager
-        # Sets sane `nix.*` defaults. Please refer to implementation/readme for more details.
-        utils.nixosModules.saneFlakeDefaults
-        (import ./modules)
-        {
-          home-manager.useGlobalPkgs = true;
-          home-manager.useUserPackages = true;
-        }
-      ];
 
       # Shared overlays between channels, gets applied to all `channels.<name>.input`
       sharedOverlays = [
--- a/examples/fully-featured/flake.nix
+++ b/examples/fully-featured/flake.nix
@@ -133,6 +133,7 @@
       checksBuilder = channels: { check = channels.nixpkgs.runCommandNoCC "test" { } "echo test > $out"; };
 
       # All other values gets passed down to the flake
+      checks.x86_64-linux.merge-with-checksBuilder-test = self.pkgs.x86_64-linux.nixpkgs.hello;
       overlay = import ./overlays;
       abc = 132;
       # etc
--- a/systemFlake.nix
+++ b/systemFlake.nix
@@ -124,29 +124,35 @@ let
   );
 
 in
-otherArguments
+mergeAny otherArguments (
 
-// eachSystem supportedSystems (system:
-  let
-    importChannel = name: value: import (patchChannel system value.input (value.patches or [ ])) {
-      inherit system;
-      overlays = sharedOverlays ++ (if (value ? overlaysBuilder) then (value.overlaysBuilder pkgs) else [ ]);
-      config = channelsConfig // (value.config or { });
-    };
+  eachSystem supportedSystems
+    (system:
+      let
+        importChannel = name: value: import (patchChannel system value.input (value.patches or [ ])) {
+          inherit system;
+          overlays = sharedOverlays ++ (if (value ? overlaysBuilder) then (value.overlaysBuilder pkgs) else [ ]);
+          config = channelsConfig // (value.config or { });
+        };
 
-    pkgs = mapAttrs importChannel channels;
+        pkgs = mapAttrs importChannel channels;
 
-    optional = check: optionalAttrs (check != null);
-  in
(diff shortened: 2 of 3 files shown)


systemFlake.nix
```diff
@@ -26,14 +29,14 @@
...
-  validateHost =
-    { channelName ? null
-    , system ? null
-    , output ? null
-    , builder ? null
+  evalHostArgs =
+    { channelName ? "nixpkgs"
+    , system ? "x86_64-linux"
+    , output ? "nixosConfigurations"
+    , builder ? channels.${channelName}.input.lib.nixosSystem
```
Pacman99, line 39: @blaggacao Would you be ok with this? Don't do any merging here, just validate and set defaults. If not I can just use the new merge function for folding hosts and merging outputs.

For this to work, we have to deprecate sharedModules and sharedExtraArgs. Which I think @gytis-ivaskevicius was ok with too. It makes the code simpler and cleans up api.

systemFlake.nix
```diff
@@ -9,11 +9,14 @@
...
 , nixosProfiles ? { } # will be deprecated soon, use hosts, instead.
 , hosts ? nixosProfiles
-, sharedExtraArgs ? { }
-, sharedModules ? [ ]
+, sharedExtraArgs ? { } # deprecate soon, prefer hostDefaults
+, sharedModules ? [ ] # deprecate soon, prefer hostDefaults
+, hostDefaults ? {
+    modules = sharedModules;
+    extraArgs = sharedExtraArgs;
+  }
```
Pacman99, line 19: So if you set hostDefaults, you can't use sharedModules. I think this is good deprecation policy.

systemFlake.nix
```diff
@@ -66,13 +69,19 @@ let
...
+        if isAttrs value then lhs.${name} or { } // value
+        else if isList value then lhs.${name} or [ ] ++ value
+        else value
+      )
+      rhs;
+
+  foldHosts = foldl' mergeAny { };
+
+  mergeArgs = lhs: rhs:
+    lhs // mapAttrs (name: value: if isAttrs value then mergeAny lhs.${name} or { } value else value) rhs;
```
Pacman99, line 84: I was thinking of merging `otherArguments` and the rest automatically, by checking if things are an attrset. It would make the below code much simpler, no need to call mergeAny on each builder expression

systemFlake.nix
```diff
@@ -66,13 +69,19 @@ let
...
-    (lhs: rhs:
-      let output = rhs.name; in
-      # manually merge whats inside the output to prevent // override
-      lhs // { ${output} = lhs.${output} or { } // rhs.value; }
-    )
-    { };
+  mergeAny = lhs: rhs:
+    lhs // mapAttrs
+      (name: value:
+        if isAttrs value then lhs.${name} or { } // value
```
Pacman99, line 75: We could make this recursive and I think we pretty much just get a local copy of `recursiveUpdate`. I'm not yet sure if this is safe however and if there was a reason for nixpkgs lib not implementing recursiveUpdate like this. But its also not necessary, recursion is only needed for otherArguments merging, but the `mergeArgs` below might be the safer solution.

gytis-ivaskevicius (approved):
It looks great! Can't wait to merge it! :+1:

systemFlake.nix
blaggacao: Does this merge lists, according to rhe merge semantics defined in previous `mergeHosts`?

systemFlake.nix
blaggacao: This sound pretty good given the generalization you made out of the merge semantics.

blaggacao (approved):
Looks pretty solid

systemFlake.nix
blaggacao: Can you confirm this accurately replaces `mergeHosts`?

systemFlake.nix
blaggacao: I think it's ok to use the mergeAny function consistently. It's a superset.

systemFlake.nix
Pacman99: Yup I've tested it with the examples you setup - really useful thanks for setting that up - and the merging is handled well.

systemFlake.nix
```diff
@@ -145,29 +124,35 @@ let
...
-  in
-  { inherit pkgs; }
-  // optional packagesBuilder { packages = packagesBuilder pkgs; }
-  // optional defaultPackageBuilder { defaultPackage = defaultPackageBuilder pkgs; }
-  // optional appsBuilder { apps = appsBuilder pkgs; }
-  // optional defaultAppBuilder { defaultApp = defaultAppBuilder pkgs; }
-  // optional devShellBuilder { devShell = devShellBuilder pkgs; }
-  // optional checksBuilder { checks = checksBuilder pkgs; }
-)
+mergeAny otherArguments (
```
Pacman99, line 127: This isn't necessary, but it allows for the `nixosConfigurations` escape hatch. You can set hosts via that and they get merged with the ones created by fup. Also could be useful if there is ever a top-level overlaysBuilder.

Either way its perfectly safe and I thought it would just be good to go ahead and merge otherArguments with fups code properly.

systemFlake.nix
```diff
@@ -145,29 +124,35 @@ let
...
+        };
+
+      in
+      { inherit pkgs; }
+      // mkOutput "packages" "packagesBuilder"
+      // mkOutput "defaultPackage" "defaultPackageBuilder"
+      // mkOutput "apps" "appsBuilder"
+      // mkOutput "defaultApp" "defaultAppBuilder"
+      // mkOutput "devShell" "devShellBuilder"
+      // mkOutput "checks" "checksBuilder"
```
Pacman99, line 152: credit to @gytis-ivaskevicius for this idea. This looks really nice and I don't have to make mergeAny recursive.

examples/fully-featured/flake.nix
```diff
@@ -36,26 +36,21 @@
...
-        # Default modules to be passed to all hosts. Equivalent to sharedModules (additive merge).
-        modules = [ ];
-      };
+        # Extra arguments to be passed to modules. Merged with host's extraArgs
+        extraArgs = { inherit utils inputs; foo = "foo"; };
+        # Default modules to be passed to all hosts.
+        modules = [
+          home-manager.nixosModules.home-manager
+          # Sets sane `nix.*` defaults. Please refer to implementation/readme for more details.
+          utils.nixosModules.saneFlakeDefaults
```
gytis-ivaskevicius, line 45: Not related to this PR - but maybe this should be a flake option like `setupFlakeRegistry = true;` or something like that

gytis-ivaskevicius (approved):
That's a lot of changes, are you sure that everything works as expected?

Pacman99:
>That's a lot of changes, are you sure that everything works as expected?

I tested hostDefaults.modules and hostDefaults.extraArgs with the examples. The resulting config of each host came out right. Folding hosts works, so theres no regression there. The outputs and builders work as expected. I have not tested patching, but this PR doesn't affect that. I think I've covered it all, but it might be good for someone else to go over it too.

examples/fully-featured/flake.nix
Pacman99: I think the current method is better, it follows the standard flake/fup api.

PR #34, opened 2021-04-16
add: specialArgs

blaggacao:
https://github.com/gytis-ivaskevicius/flake-utils-plus/pull/26/files apparently never made it into staging for a merging error of mine, I guess.

### DIFF
--- a/systemFlake.nix
+++ b/systemFlake.nix
@@ -40,6 +40,7 @@ let
     , builder ? channels.${channelName}.input.lib.nixosSystem
     , modules ? [ ]
     , extraArgs ? { }
+    # These are not part of the module system, so they can be used in `imports` lines without infinite recursion
     , specialArgs ? { }
     }: { inherit channelName system output builder modules extraArgs specialArgs; };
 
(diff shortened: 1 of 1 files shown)


systemFlake.nix
```diff
@@ -40,7 +40,8 @@ let
     , builder ? channels.${channelName}.input.lib.nixosSystem
     , modules ? [ ]
     , extraArgs ? { }
-    }: { inherit channelName system output builder modules extraArgs; };
+    , specialArgs ? { }
```
Pacman99, line 43:
```suggestion
    # These are not part of the module system, so they can be used in `imports` lines without infinite recursion
    , specialArgs ? { }
```
Just to explain the difference, so others don't have to hunt down comments in lib/modules.nix

blaggacao:
Merging this as a fixup of previous merge mistakes (and we need to to proceed with `devos` rebase).

local nixpkgs-fmt yields, I guess we have to provide a devshell so versions can be properly pinned:
```
➜  flake-utils-plus git:(da/specialArgs-back-again) nixpkgs-fmt --check .
0 / 13 would have been reformatted
```

PR #35, opened 2021-04-16
add simple devshell

blaggacao:
```console
$ flake-utils-plus git:(da/add-devshell) nix develop
🔨 Welcome to flake-utils-plus

[checks]

  check-fully-featured-example
  check-somewhat-realistic-example
  evalnix

[formatters]

  fmt

[general commands]

  menu                             - prints this menu

[flake-utils-plus]$
```

---

I also in-lined the CI with this approach &mdash; and surfaced a bug, I guess.

Please consider commit messages, somewhat individually.


Pacman99:
I'm not sure about this. I don't think devshell is really needed for fup development. But also this does sort of add a nixpkgs/devshell dependency, but in a weird way. It feels like fup would now be unofficially dependent on nixpkgs and devshell. Almost like false advertising. 
The decision has been made to use pure nix and if we're trying to add features or tools to the project that require an additional dependency, then we need to evaluate if fup really should be adding that feature or tool.

blaggacao:
The issue that really warranted this is that my local version of `nixpkgs-fmt` formatted differently than the one used by CI. The only solution is to pin nixpkgs for that purpose. Somehow.

I was also frustrated, that it wasn't cheap (enough) to locally test stuff against the examples.

Devshell is a match made in heaven to comfortably streamline CI & dev environment. It has thise nice MOTD feature and makes it obvious and cheaply discoverable for contributors to preemt CI checks, locally.

Many dependency systems provide for dev dependencies that are completely optional. Flake's don't seem to have gained that distinction yet, and we don't want to overload normal users of this flake with dependencies that are completely irrelevant for the library.

The work-around as exposed here is copied over from the devshell repo, which does basically the same thing. It's the only way I've found to decouple those deps &mdash; a part from implementing a subflake for shell/ci, but that would break the increasingly well known `nix develop` workflow at the expense of doing something like nix develop ./shell-sub-flake or so.

blaggacao:
I'll finish this off tomorrow early on.

Pacman99:
I get all the advantages devshell provides, so I'm not entirely against this. I'm just worried about the fetchTarball calls, it feels a bit wrong and "hides" dependencies of fup. Development dependencies can still be considered dependencies.

> The work-around as exposed here is copied over from the numtide/devshell repo, which does basically the same thing. It's the only way I've found to decouple those deps — apart from implementing a subflake for shell/ci, but that would break the increasingly well known nix develop workflow at the expense of doing something like nix develop ./shell-sub-flake or so.

I think I would actually prefer the sub flake. That way we use standard flake apis - I'm against fetchTarball when flakes exist - and a user can see exactly what the development dependencies are instead of them being hidden in a shell.nix. And it won't pollute the library's dependencies.

blaggacao:
> I think I would actually prefer the sub flake.

I completely hear where you are aiming at. And I share you're principled approach.

It would have a couple of consequences, though:
- We'd have to stack it away in a subfolder
- We'd have to deviate from `nix develop` (the common standard to hack on a repo). If we "pull up" the entire sub-flake, we also pull up their entire dependencies, also for users.
- It would not work for people that still hack on `nix-shell`

shell.nix
```diff
@@ -0,0 +1,21 @@
...
+
+  nixpkgsSrc = fetchTarball {
+    url = "https://github.com/NixOS/nixpkgs/archive/${nixpkgsGitRev}.tar.gz";
+    sha256 = "080fvmg0i6z01h6adddfrjp1bbbjhhqk32ks6ch9gv689645ccfq";
+  };
+
+  devshellSrc = fetchTarball {
+    url = "https://github.com/numtide/devshell/archive/${devshellGitRev}.tar.gz";
+    sha256 = "1px9cqfshfqs1b7ypyxch3s3ymr4xgycy1krrcg7b97rmmszvsqr";
+  };
```
Pacman99, line 15: Is it ok to call fetchTarball in pure mode? I thought even with passing the hash its still not allowed.

Pacman99 (approved):
I read through this a bit more and I think I'm more okay with it. It improves checks a lot - which I clearly need - and it looks like it would be useful for development.
The devshell looks really nice! I like the new checks.

shell.nix
blaggacao: Maybe the fact that `nix develop` is working implicitly answers this with "yes"? I think I remember ix 3.0 is pure by default? Not sure at all, though.

.github/workflows/ci.yaml
```diff
@@ -9,7 +9,7 @@ jobs:
       - uses: cachix/install-nix-action@v13
       - name: Check Nix parsing
         run: |
-          find . -name "*.nix" -exec nix-instantiate --parse --quiet {} >/dev/null +
+          nix-shell --run evalnix
```
gytis-ivaskevicius, line 12: `nix-shell` - so much for migration to flakes :D Can we use nix 3.0 syntax, please?

shell.nix
gytis-ivaskevicius: fetch is fine as long as there is a sha256 specified

gytis-ivaskevicius (changes requested):
Okay, so first of all action ci.yaml is outdated, as for devshell thingy - I'd say an obvious solution is to implement it as part of one of the examples or as a separate flake on its own. This way we would not need to care about updating nixpkgs/devshell hashes

Another solution is to evaluate shell.nix manually

Anyways, the issue seems to exist due to the version difference on stable/unstable channels (1.0.0 vs 1.2.0). Pinning it locally is an option as well


If current implementation is fine by both of you - it is fine by me as well

Pacman99:
I'm not yet convinced on why it can't be put into a subfolder.
`nix develop ./shell` isn't that bad. We can still have a root level shell.nix to support `nix-shell` users - not that many fup developers would be using nix-shell.
github actions is a good argument against this. But could we setup cachix to use nix-shell instead? and put a shell.nix in root level to work with that.

gytis-ivaskevicius:
So, any changes planned? should we fix up conflicts and merge this one as well?

blaggacao:
I can rebase, but I have to let go on further improvements.

If we can merge that then, I shall be hugely rewarded.

blaggacao:
Final shape, good to go:
```console
$ 🔨 Welcome to flake-utils-plus

[checks]

  evalnix                    - Check Nix parsing
  fmt                        - Check Nix formatting

[checks (fully featured example)]

  nix-build-morty-ff         - Build Morty configuration
  nix-build-rick-ff          - Build Rick configuration
  nix-build-summer-ff        - Build Summer configuration
  nix-flake-check-ff         - Run 'nix flake check'
  nix-flake-show-ff          - Run 'nix flake show'

[checks (somewhat realistic example)]

  nix-build-hostnameone-sr   - Build HostnameThree configuration
  nix-build-hostnamethree-sr - Build HostnameThree (darwin) configuration
  nix-flake-check-sr         - Run 'nix flake check'
  nix-flake-show-sr          - Run 'nix flake show'

[general commands]

  menu                       - prints this menu

[flake-utils-plus]$
```

blaggacao:
@gytis-ivaskevicius rebase-merge? (it's just dev tooling convenience)

gytis-ivaskevicius (approved):
Looks fine to me. Rebase is probably unnecessary. Feel free to merge it :+1: 

Also, here is another solution to this devshell dependency issue (might be cool to do so in the long run): 
Ripoff another numtide project and add it as part of this repo :D

blaggacao:
> Also, here is another solution to this devshell dependency issue

I'm convinced the only correct solution is ro add the notion of development dependencies (not build dependencies).

The nix ecosystem has not yet "outgrown" the build(-only) focus, of which the original idwa of `nix-shell` to give access to the _build_ instead of the _development_ environment still testifies.

PR #36, opened 2021-04-17
Use nixos modules and lib from patched nixpkgs and allow setting nixpkgs.config values in host modules

### LINKED ISSUE #4: Setting nixpkgs.config in a module doesn't seem to work anymore
I have some infrastructure in my nixos config to allow only specific unfree packages (to avoid accidental unfree use). Therefore I need to set `nixpkgs.config.allowUnfreePredicate` in my config, but the value it is set to doesn't seem to work anymore.

chvp:
As far as I can tell, these changes allow patching NixOS and configuring `nixpkgs.config` in the host's modules. I was able to evaluate (and switch to) my nixos config which uses both correctly.

I'll clean this up some more and maybe demonstrate the different possibilities in the fully featured example flake. (That will have to wait until tomorrow though, it's quite late here.)

Fixes #4.

### DIFF
--- a/examples/fully-featured/flake.nix
+++ b/examples/fully-featured/flake.nix
@@ -116,6 +116,10 @@
           output = "someConfigurations";
         };
 
+        Summer = {
+          channelName = "unstable";
+          modules = [ ./configurations/Summer.host.nix ];
+        };
       };
 
       # Evaluates to `packages.<system>.attributeKey = "attributeValue"`
@@ -134,7 +138,19 @@
       devShellBuilder = channels: channels.nixpkgs.mkShell { name = "devShell"; };
 
       # Evaluates to `checks.<system>.attributeKey = "attributeValue"`
-      checksBuilder = channels: { check = channels.nixpkgs.runCommandNoCC "test" { } "echo test > $out"; };
+      checksBuilder = channels:
+        let
+          booleanCheck = cond:
+            if cond
+            then channels.nixpkgs.runCommandNoCC "success" { } "echo success > $out"
+            else channels.nixpkgs.runCommandNoCC "failure" { } "exit 1";
+        in
+        {
+          check = channels.nixpkgs.runCommandNoCC "test" { } "echo test > $out";
+          summerHasCustomModuleConfigured = booleanCheck (self.nixosConfigurations.Summer.config.patchedModule.test == "test");
+          summerHasPackageOverridesConfigured = booleanCheck (self.nixosConfigurations.Summer.config.nixpkgs.pkgs.config ? packageOverrides);
+          summerHasUnfreeConfigured = booleanCheck (self.nixosConfigurations.Summer.config.nixpkgs.pkgs.config ? allowUnfree);
+        };
 
       # All other values gets passed down to the flake
       checks.x86_64-linux.merge-with-checksBuilder-test = self.pkgs.x86_64-linux.nixpkgs.hello;
--- a/examples/fully-featured/flake.nix
+++ b/examples/fully-featured/flake.nix
@@ -147,8 +147,11 @@
         in
         {
           check = channels.nixpkgs.runCommandNoCC "test" { } "echo test > $out";
+          # Modules (and lib) from patched nixpkgs are used
           summerHasCustomModuleConfigured = booleanCheck (self.nixosConfigurations.Summer.config.patchedModule.test == "test");
+          # nixpkgs config from host-specific module is used
           summerHasPackageOverridesConfigured = booleanCheck (self.nixosConfigurations.Summer.config.nixpkgs.pkgs.config ? packageOverrides);
+          # nixpkgs config from channel is also used
           summerHasUnfreeConfigured = booleanCheck (self.nixosConfigurations.Summer.config.nixpkgs.pkgs.config ? allowUnfree);
         };
 
--- a/systemFlake.nix
+++ b/systemFlake.nix
@@ -86,12 +86,14 @@ let
     let
       selectedNixpkgs = getNixpkgs host;
       host = evalHostArgs (mergeAny hostDefaults host');
+      channelValue = channels.${host.channelName};
+      patchedChannel = patchChannel host.system channelValue.input (channelValue.patches or [ ]);
     in
     {
       ${host.output}.${hostname} = host.builder {
-        inherit (selectedNixpkgs) system;
+        inherit (host) system;
         modules = [
-          ({ pkgs, lib, options, ... }: {
+          ({ pkgs, lib, options, config, ... }: {
             # 'mkMerge` to separate out each part into its own module
             _type = "merge";
             contents = [
@@ -99,11 +101,13 @@ let
                 networking.hostName = hostname;
               })
 
-              (if options ? nixpkgs then {
-                nixpkgs = {
-                  inherit (selectedNixpkgs) overlays config system;
+              (optionalAttrs (options ? nixpkgs) {
+                nixpkgs.pkgs = import patchedChannel {
+                  inherit (host) system;
+                  overlays = sharedOverlays ++ (if (channelValue ? overlaysBuilder) then (channelValue.overlaysBuilder pkgs) else [ ]);
+                  config = channelsConfig // (channelValue.config or { }) // config.nixpkgs.config;
                 };
-              } else { _module.args.pkgs = selectedNixpkgs; })
+              })
 
               (optionalAttrs (options ? system.configurationRevision) {
                 system.configurationRevision = lib.mkIf (self ? rev) self.rev;
--- a/systemFlake.nix
+++ b/systemFlake.nix
@@ -101,13 +101,13 @@ let
                 networking.hostName = hostname;
               })
 
-              (optionalAttrs (options ? nixpkgs) {
+              (if options ? nixpkgs then {
                 nixpkgs.pkgs = import patchedChannel {
                   inherit (host) system;
                   overlays = sharedOverlays ++ (if (channelValue ? overlaysBuilder) then (channelValue.overlaysBuilder pkgs) else [ ]);
                   config = channelsConfig // (channelValue.config or { }) // config.nixpkgs.config;
                 };
-              })
+              } else { _module.args.pkgs = selectedNixpkgs; })
 
               (optionalAttrs (options ? system.configurationRevision) {
                 system.configurationRevision = lib.mkIf (self ? rev) self.rev;
--- a/systemFlake.nix
+++ b/systemFlake.nix
@@ -88,6 +88,21 @@ let
       host = evalHostArgs (mergeAny hostDefaults host');
       channelValue = channels.${host.channelName};
       patchedChannel = patchChannel host.system channelValue.input (channelValue.patches or [ ]);
+      # Use lib from patched nixpkgs
+      lib = import (patchedChannel + "/lib");
+      # Use nixos modules from patched nixpkgs
+      baseModules = import (patchedChannel + "/nixos/modules/module-list.nix");
+      # Override `modulesPath` because otherwise imports from there will not use patched nixpkgs
+      specialArgs = { modulesPath = builtins.toString (patchedChannel + "/nixos/modules"); } // host.specialArgs;
+      # The only way to find out if a host has `nixpkgs.config` set to
+      # the non-default value is by evalling most of the config.
+      hostConfig = (lib.evalModules {
+        prefix = [ ];
+        check = false;
+        modules = baseModules ++ host.modules;
+        args = { inherit inputs; } // host.extraArgs;
+        inherit specialArgs;
+      }).config;
     in
     {
       ${host.output}.${hostname} = host.builder ({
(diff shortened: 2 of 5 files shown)


blaggacao:
Since importing `nixpkgs` unfortunately is not a cheap operation, I'd propose to condition any such logic on the presence of `nixpkgs.config` and otherwise share the (already) instantiated `self.pkgs` betwewn hosts.

chvp:
> Since importing `nixpkgs` unfortunately is not a particular cheap operation, I'd propose to condition any such logic on the meaningful presence of `nixpkgs.config` (whatever that would be) and otherwise stick with sharing the (already) instantiated per-channel-per-system `self.pkgs` between hosts.

Figuring out the presence of `nixpkgs.config` is not an easy operation without triggering infinite recursion, sadly. As far as I can tell, it would require an extra evaluation of the host's configured modules. I'd have to try it to see if it is even possible at all.

> Other than that, this looks like a pretty nice escape hatch to avoid rendering the nix manual deprecated. 😉

Not sure what you mean by this?

chvp:
The current iteration makes it so that the nixpkgs patching can also patch `lib` and `nixos/modules`. I'm pretty sure it doesn't result in an extra nixpkgs evaluation, since `selectedNixpkgs` is only used if the `nixpkgs.config` option does not exist.

This probably breaks the `nix-darwin` support right now, but I'm not sure how I can test that.

blaggacao:
For the patching, we really should see if we (as a community of fup / devos users) can give https://github.com/NixOS/nix/issues/3920#issuecomment-806622462 a spin. Wouldn't you (or anybody reading this) know of somebody capable (in C) of getting the work done?

systemFlake.nix
```diff
@@ -100,8 +102,10 @@ let
               })
 
               (if options ? nixpkgs then {
-                nixpkgs = {
-                  inherit (selectedNixpkgs) overlays config system;
+                nixpkgs.pkgs = import patchedChannel {
+                  inherit (host) system;
+                  overlays = sharedOverlays ++ (if (channelValue ? overlaysBuilder) then (channelValue.overlaysBuilder self.pkgs) else [ ]);
+                  config = channelsConfig // (channelValue.config or { }) // config.nixpkgs.config;
```
blaggacao, line 108: @Pacman99 The condensed reason for this is `config.nixpkgs.config` to work as the nix-manual states.

However it odes import nixpkgs on every host and does not "re-use" the import of `self.pkgs`. Conceptually this is fine, since anything host-specific would have no natuarl pace in `self.pkgs` (which is system / channel specific).

However, I'm a bit vary that we would import nixpkgs for every single host (again). That doesn't scale very well if we have 5000 otherwise uniform hosts to update, for example.

I'd be very much in favor of this addition if it was possible to only trigger it (and opt to not re-use imports from `self.pkgs`) in case `config.nixpkgs.config` was meaningfully set.

I ping you since, maybe you might have an idea here how to accomplish this.

chvp:
> For the patching, we really should see if we (as a community of fup / devos users) can give [NixOS/nix#3920 (comment)](https://github.com/NixOS/nix/issues/3920#issuecomment-806622462) a spin. Wouldn't you (or anybody reading this) know of somebody capable (in C) of getting the work done?

I don't, sorry.

systemFlake.nix
chvp: Note that the `self.pkgs` import never actually happens, since Nix is lazy, and `self.pkgs` isn't used anywhere for a nixos configuration.

Like I said, detecting use of nixpkgs.config is not easy. The full configuration for a host would have to be evaluated separately to avoid infinite recursion (we can't check whether something is set in what we are currently evaluating if we might change the outcome of that check based on the result of that check).

chvp:
The last commit adds the proposed mechanism to avoid re-importing nixpkgs for a host unless the host sets `nixpkgs.config`.

blaggacao (approved):
LGTM, reviewed on cel phone.

systemFlake.nix
```diff
@@ -86,24 +86,51 @@ let
...
       selectedNixpkgs = getNixpkgs host;
       host = evalHostArgs (mergeAny hostDefaults host');
+      channelValue = channels.${host.channelName};
+      patchedChannel = patchChannel host.system channelValue.input (channelValue.patches or [ ]);
+      # Use lib from patched nixpkgs
+      lib = import (patchedChannel + "/lib");
+      # Use nixos modules from patched nixpkgs
+      baseModules = import (patchedChannel + "/nixos/modules/module-list.nix");
+      # Override `modulesPath` because otherwise imports from there will not use patched nixpkgs
+      specialArgs = { modulesPath = builtins.toString (patchedChannel + "/nixos/modules"); } // host.specialArgs;
```
gytis-ivaskevicius, line 96: I believe using something along the lines of `patchedChannel = selectedNixpkgs.path;` should get the job done nicely.
Same goes for lib - `lib = selectedNixpkgs.lib;`
As for `baseModules` - I bet there is some way to do it as well, need to look into it. I will get back to you on this one.

systemFlake.nix
```diff
@@ -86,24 +86,51 @@ let
...
+              # because that triggers infinite recursion.
+                if (hostConfig.nixpkgs.config == { }) then
+                  { nixpkgs.pkgs = selectedNixpkgs; }
+                else
+                  {
+                    nixpkgs.pkgs = import patchedChannel {
+                      inherit (host) system;
+                      overlays = sharedOverlays ++ (if (channelValue ? overlaysBuilder) then (channelValue.overlaysBuilder self.pkgs) else [ ]);
+                      config = channelsConfig // (channelValue.config or { }) // config.nixpkgs.config;
+                    };
```
gytis-ivaskevicius, line 131: Consider this instead:
```nix
                {
                  nixpkgs.pkgs =
                    if (hostConfig.nixpkgs.config == { })
                    then selectedNixpkgs
                    else
                      import patchedChannel {
                        inherit (host) system;
                        overlays = sharedOverlays ++ (if (channelValue ? overlaysBuilder) then (channelValue.overlaysBuilder self.pkgs) else [ ]);
                        config = channelsConfig // (channelValue.config or { }) // config.nixpkgs.config;
                      };
                }
```

systemFlake.nix
```diff
@@ -122,7 +149,9 @@ let
           })
         ] ++ host.modules;
         specialArgs = host.specialArgs;
-      };
+      } // (optionalAttrs (host.output == "nixosConfigurations") {
```
gytis-ivaskevicius, line 152: I guess such comparison is fine, but I'd love to hear if anyone else has better ideas to detect nixos system

systemFlake.nix
gytis-ivaskevicius: Wait a second, won't this use a patched version of nixpkgs in case no configuration defined at all? :thinking:

gytis-ivaskevicius (approved):
Aside from these minor changes - it really does look great, very nice!!

One of the things that I'd like to ask is to add some sort of "tests" to make sure that we don't break this functionality in the future. This can be done via just using these changes in one of the examples and making sure that pipelines pass https://github.com/chvp/flake-utils-plus/actions 

It's like either you do it or I will :D, I think it's important since to test something like that is a little tricky. No worries if you rather not deal with it.


Anyways, fixup or shoot down these comments and send a message "Ready to be merged" and it's going to be the end of this PR :)

systemFlake.nix
Pacman99: I was thinking of a possible alternative to this could be if you set the builder to `import "${selectedNixpkgs}/nixos/lib/eval-config.nix"`. That would only deal with the modules, but it wouldn't help with the nixpkgs.config issue.

lib.nixosSystem is essentially that but it sets the nixos revision and adds some vm features. So if we are willing to drop those, and just directly import eval-config, it might be much cleaner.

Pacman99: Or maybe we can keep the default as nixosSystem, and just add documentation on how to pass the patched channel eval-config as the builder.

systemFlake.nix
Pacman99: I think it would be better if we this was done with the custom builder argument.

systemFlake.nix
chvp: Both use patched nixpkgs, the difference is the config that is passed to the patched nixpkgs. In the first case the config is only the config set in flake.nix, in the second the configuration from the host's modules is taken into account as well.

systemFlake.nix
chvp: Ah, I didn't know the path attribute existed. I'll make the requested changes. I'm pretty sure the base modules will always have to be imported, they aren't exposed in nixpkgs' pkgs.

systemFlake.nix
chvp: Moving this to the builder seems like a good idea. I'd argue for it to be the default, though.

chvp:
> Aside from these minor changes - it really does look great, very nice!!
> 
> One of the things that I'd like to ask is to add some sort of "tests" to make sure that we don't break this functionality in the future. This can be done via just using these changes in one of the examples and making sure that pipelines pass https://github.com/chvp/flake-utils-plus/actions
> 
> It's like either you do it or I will :D, I think it's important since to test something like that is a little tricky. No worries if you rather not deal with it.
> 
> Anyways, fixup or shoot down these comments and send a message "Ready to be merged" and it's going to be the end of this PR :)

I'll add some tests tomorrow. It's something I definitely wanted to do, but I was waiting to see whether you thought this PR was viable or not.

systemFlake.nix
gytis-ivaskevicius: I feel like a custom builder would make it more complex than it needs to be for the end-user, maybe we should keep current implementation?

chvp:
I added some test that check whether the patched modules and lib is used and whether the config is merged correctly when the host has `nixpkgs.config` set. The pipeline passes except for the last item (the Darwin configuration). https://github.com/chvp/flake-utils-plus/runs/2380823370

examples/fully-featured/flake.nix
```diff
@@ -134,7 +138,22 @@
...
+        in
+        {
+          check = channels.nixpkgs.runCommandNoCC "test" { } "echo test > $out";
+          # Modules (and lib) from patched nixpkgs are used
+          summerHasCustomModuleConfigured = booleanCheck (self.nixosConfigurations.Summer.config.patchedModule.test == "test");
+          # nixpkgs config from host-specific module is used
+          summerHasPackageOverridesConfigured = booleanCheck (self.nixosConfigurations.Summer.config.nixpkgs.pkgs.config ? packageOverrides);
+          # nixpkgs config from channel is also used
+          summerHasUnfreeConfigured = booleanCheck (self.nixosConfigurations.Summer.config.nixpkgs.pkgs.config ? allowUnfree);
+        };
```
gytis-ivaskevicius, line 156: Alright, you put more work into it than expected - I would have just set some host-specific configuration and checked if it builds.

I like the fact that it's a check but I am not sure if this should be part of the example. But then again - this example is already pretty overwhelming.

@Pacman99 @blaggacao Any thoughts?

systemFlake.nix
```diff
@@ -86,24 +86,51 @@ let
...
       host = evalHostArgs (mergeAny hostDefaults host');
+      patchedChannel = selectedNixpkgs.path;
+      # Use lib from patched nixpkgs
+      lib = selectedNixpkgs.lib;
+      # Use nixos modules from patched nixpkgs
+      baseModules = import (patchedChannel + "/nixos/modules/module-list.nix");
+      # Override `modulesPath` because otherwise imports from there will not use patched nixpkgs
+      specialArgs = { modulesPath = builtins.toString (patchedChannel + "/nixos/modules"); } // host.specialArgs;
+      # The only way to find out if a host has `nixpkgs.config` set to
+      # the non-default value is by evalling most of the config.
```
gytis-ivaskevicius, line 97: I'm quite happy with it. I'll take a closer look into those other two cases at... at... at some point :D
But yeah, LGTM

gytis-ivaskevicius (approved):
Looks great, will wait for a little while to see if @Pacman99 or @blaggacao has any comments to add and I guess we might as well merge it :)

Also, if you don't mind - mind rebasing this branch on the latest staging?

chvp:
> Also, if you don't mind - mind rebasing this branch on the latest staging?

Done

examples/fully-featured/flake.nix
blaggacao: IMO, it's completely ok to have the fulky featured example overflow. Where can we have such detailed documentation that is not called "tests", otherwise? :wink:

systemFlake.nix
blaggacao: We definitely need to find somebody who knows how to implement `inputs.myflake.patches = [];` :smile:

blaggacao (approved):
Still LGTM! Thanks for the in-line comments.

chvp:
There was still a problem with nix-darwin evaluation (which the rebase caught), so I fixed that in the latest commit.

Pacman99 (approved):
LGTM I'm not a huge fan of the output == nixosConfigurations, but also I don't see any way around it. I like that this helps with improving patching, while not hurting workflows that don't use it.

Also the creation of `hostConfig` is a really cool trick with evalModules that I've never been able to get right, and I could see it being useful in other places.

PR #40, opened 2021-04-18
removes unnecessary newlines

legendofmiracles:


examples/fully-featured-flake.nix
```diff
@@ -133,7 +133,3 @@
 
     };
 }
-
```
gytis-ivaskevicius, line 136: Bro, but where is that extra newline at the end? :D

gytis-ivaskevicius (changes requested):
All the development is happening in the staging branch. This change is incompatable

PR #42, opened 2021-04-22
imp: provide per channel modules paths to make it easy to "backport"

blaggacao:
a module. imports = [ ${myChannelModulesPathk}/installer/...

This allows users to relatively conveniently pull in modules from other channels. The chances of success are mixed depending on the channel distance, but if it works, it's great.

### DIFF
--- a/systemFlake.nix
+++ b/systemFlake.nix
@@ -30,7 +30,7 @@
 
 let
   inherit (flake-utils-plus.lib) eachSystem patchChannel;
-  inherit (builtins) foldl' mapAttrs removeAttrs attrValues isAttrs isList;
+  inherit (builtins) foldl' mapAttrs removeAttrs attrValues attrNames isAttrs isList;
 
   # set defaults and validate host arguments
   evalHostArgs =
@@ -42,7 +42,9 @@ let
     , extraArgs ? { }
       # These are not part of the module system, so they can be used in `imports` lines without infinite recursion
     , specialArgs ? { }
-    }: { inherit channelName system output builder modules extraArgs specialArgs; };
+    }: {
+      inherit channelName system output builder modules extraArgs specialArgs;
+    };
 
   # recursively merge attribute sets and lists up to a certain depth
   mergeAny = lhs: rhs:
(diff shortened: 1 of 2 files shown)


Pacman99 (approved):
I like this, it just adds more information to make backporting/updating modules possible.

I think this should go inside `configurationBuilder` instead and possibly under the `optionalAttrs (output == "nixosConfigurations")`. It is a default, so its not terrible to put in evalHostArgs, but also its nixos specific and theres another place where modulesPath is edited, so it makes sense to go there.

systemFlake.nix
```diff
@@ -42,7 +42,16 @@ let
...
       # These are not part of the module system, so they can be used in `imports` lines without infinite recursion
     , specialArgs ? { }
-    }: { inherit channelName system output builder modules extraArgs specialArgs; };
+    }: {
+      inherit channelName system output builder modules extraArgs;
+      specialArgs = let
+        f = channelName:
+          { "${channelName}ModulesPath" = "${channels.${channelName}.input}/nixos/modules"; };
+        channelModulesPath = map f (attrNames channels);
+        merge = lhs: rhs: lhs // rhs;
```
gytis-ivaskevicius, line 51: I don't think there is much of a point for this merge function :D

gytis-ivaskevicius (approved):
Looks good to me, tho let's not merge it just yet. I was about to fix existing issues in staging and actually do something about the release since we really need one

systemFlake.nix
```diff
@@ -149,9 +159,9 @@ let
             ];
           })
         ] ++ host.modules;
-        specialArgs = host.specialArgs;
+        inherit specialArgs;
```
blaggacao, line 162: this was in failure mode before. we have to remain careful about the builder generalization.

Pacman99: There was actually a reason for this: setting modulesPath on a non-nixos builder to a nixos path would break things.
I'll address this in #44

blaggacao: Ok thanks! We have to be really careful about rhose kind of things. There are now +- 5 design dimenstion in three lines of code.

blaggacao: Maybe we'd need to split them eventually apart completely and implement a generic builder and a specific nixos one on top.

PR #43, opened 2021-04-25
General exporter and builderConstructor improvements

Pacman99:
This is some fixups and improvments to the api for the overlaysFromChannelsExporter and packagesFromOverlaysBuilderConstructor. 

Have overlays exporter take `self.pkgs` and `inputs`. The former is so nixpkgs sets are instantiated and the overlay lists are ready. The latter is to auto filter any overlays that were taken from inputs. Its not good to export things that weren't originated in your flake and you don't plan on maintaining.
 
Change packages builderConstructor to just take an attrset of overlays and the builder can then pull out packages defined in overlays from the channels set.

### DIFF
--- a/overlaysFromChannelsExporter.nix
+++ b/overlaysFromChannelsExporter.nix
@@ -1,11 +1,13 @@
 { flake-utils-plus }:
 let
 
-  overlaysFromChannelsExporter = channels:
+  overlaysFromChannelsExporter = { pkgs, inputs ? { } }:
     /**
-      Synopsis: overlaysFromChannelsExporter _channels_
+      Synopsis: overlaysFromChannelsExporter _{ pkgs, inputs }_
 
-      channels: channels.<name>.overlays
+      pkgs: self.pkgs
+
+      inputs: flake inputs to sort out external overlays
 
       Returns an attribute set of all packages defined in an overlay by any channel
       intended to be passed to be exported via _self.overlays_. This method of
@@ -21,6 +23,11 @@ let
       In the case of the unstable channel, this information is still of varying usefulness,
       as effective cut dates can vary heavily between repositories.
 
+      To ensure only overlays that originate from the flake are exported you can optionally pass
+      a set of flake inputs and any overlay which is taken from an input will be filtered out.
+      Optimally this would be done by detecting flake ownership of each overlay, but that is not 
+      possible yet, so this is the next best workaround.
+
       Example:
 
       overlays = [
@@ -31,13 +38,22 @@ let
 
       **/
     let
-      inherit (builtins) mapAttrs attrNames concatMap listToAttrs;
+      inherit (builtins) mapAttrs foldl' filter head attrNames attrValues concatMap listToAttrs elem;
       nameValuePair = name: value: { inherit name value; };
+
+      # just pull out one arch from the system-spaced pkgs to get access to channels
+      # overlays can be safely evaluated on any arch
+      channels = head (attrValues pkgs);
 
       pathStr = path: builtins.concatStringsSep "/" path;
 
       channelNames = attrNames channels;
       overlayNames = overlay: attrNames (overlay null null);
+
+      # get all overlays from inputs
+      inputOverlays = mapAttrs (_: v: [ v.overlay or (_: _: { }) ] ++ attrValues v.overlays or { }) inputs;
+      # use overlayNames as a way to identify overlays
+      flattenedInputOverlays = map overlayNames (foldl' (a: b: a ++ b) [ ] (attrValues inputOverlays));
 
       extractAndNamespaceEachOverlay = channelName: overlay:
         map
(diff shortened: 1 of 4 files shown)


Pacman99:
Not sure whats up with the nixpkgs-fmt check, I ran it on my system and the file doesn't get changed.

Pacman99:
This depends on https://github.com/numtide/flake-utils/pull/31

packagesFromOverlaysBuilderConstructor.nix
```diff
@@ -1,89 +1,76 @@
 { flake-utils-plus }:
 let
 
-  overlayFromPackagesBuilderConstructor = channels:
+  overlayFromPackagesBuilderConstructor = overlays:
```
blaggacao, line 4: If you don't mind, would you cover up my misnomer? It should be `packagesFromOverlaysBuilderConstructor`. I probably was convincingly conaitent woth that error down the file.

Sry!

:wink:

packagesFromOverlaysBuilderConstructor.nix
```diff
@@ -1,89 +1,76 @@
...
-          "development/vim" = { ... };
-          };
-
-          overlays' = [
-          "development"
-          ];
-
-          overlays' will pass both pkgs' through the sieve.
+            self.packages = {
+              firefox = prev.override { privacySupport = true; };
```
blaggacao, line 33: I reckon technically, it couldn't be `prev.`, I think you wanted to put in a placeholder that represents a `<derivation-with-overlay-applied>`

overlaysFromChannelsExporter.nix
```diff
@@ -31,14 +33,23 @@ let
 
       **/
     let
-      inherit (builtins) mapAttrs attrNames concatMap listToAttrs;
+      inherit (builtins) mapAttrs foldl' filter attrNames attrValues concatMap listToAttrs elem;
       nameValuePair = name: value: { inherit name value; };
 
+      # overlays aren't system specific so arch doesn't matter here
+      channels = pkgs.x86_64-linux;
```
blaggacao, line 40: Could we rather use the first element of whatever aarch is defined there? That would appear to me as a safer way which could not enter the failure mode where x86 is not among the user-supported architecture.

blaggacao (approved):
As an extremely promising API fix, this should make it into the upcoming release.

packagesFromOverlaysBuilderConstructor.nix
Pacman99: good point, just wrote out what I was going for in text

packagesFromOverlaysBuilderConstructor.nix
Pacman99: fixed to follow the same naming

overlaysFromChannelsExporter.nix
Pacman99: I used head and attrValues to fix this and just get the first system in alphabetical order. It doesn't really matter what system it uses, just needs to get the instantiated.

Pacman99:
This should be ready to go, I just changed the commit to update flake-utils, since the fixup PR has been merged.

overlaysFromChannelsExporter.nix
```diff
@@ -31,14 +33,24 @@ let
...
+      # overlays can be safely evaluated on any arch
+      channels = head (attrValues pkgs);
+
       pathStr = path: builtins.concatStringsSep "/" path;
 
       channelNames = attrNames channels;
       overlayNames = overlay: attrNames (overlay null null);
 
+      /* workaround to detect overlays owned by this flake, by filtering out overlays from others */
+      # get all overlays from inputs
```
gytis-ivaskevicius, line 49: Whats up with different type of comments :D

packagesFromOverlaysBuilderConstructor.nix
```diff
@@ -1,92 +1,81 @@
...
-          (_: pkgs:
-            filterAttrs
-              (pkgName:
-                any (overlayName: hasPrefix pkgName overlayName) flattendOverlaysNames
-              )
-              pkgs
-          )
-          flattenedPackages;
+        # flatten any nested sets with with "/" delimiter
+          # filter out packages that don't meet the flake check conditions
```
gytis-ivaskevicius, line 74: Why doesn't nixpkgs-fmt pick up these comments alignment :/

gytis-ivaskevicius (approved):
Looks great, tho Ill need to try it out with odd values to see if it is actually working as expected.

Feel free to merge it whenever you think its ready

overlaysFromChannelsExporter.nix
Pacman99: I was going for a header, but I think I'm going to move that comment to the top next to the `inputs` argument description. So its clear why its needed.

packagesFromOverlaysBuilderConstructor.nix
Pacman99: actually `nixpkgs-fmt` forces that formatting, I tried to put them both on the same indent, but fmt kept staggering them. I condensed it to just one comment.

PR #44, opened 2021-04-25
embed info in channels and pass host's channel as specialArg

Pacman99:
gives both builders and hosts information about what channel their on. The former is so you can implement channel-specific logic in a builder. In devos we have a extended nixosSystem, so we would want to do `specialArgs.channel.input.lib.nixosSystem`, to ensure we are using the write channel for that host. And modules can do logic to hosts based on what channel the host is on.

### DIFF
--- a/systemFlake.nix
+++ b/systemFlake.nix
@@ -89,20 +89,23 @@ let
       selectedNixpkgs = getNixpkgs host;
       host = evalHostArgs (mergeAny hostDefaults host');
       patchedChannel = selectedNixpkgs.path;
+
+      specialArgs = host.specialArgs // { channel = selectedNixpkgs; };
+
+      /* nixos specific arguments */
       # Use lib from patched nixpkgs
       lib = selectedNixpkgs.lib;
       # Use nixos modules from patched nixpkgs
       baseModules = import (patchedChannel + "/nixos/modules/module-list.nix");
-      specialArgs =
+      nixosSpecialArgs =
         let
           f = channelName:
-            { "${channelName}ModulesPath" = builtins.toString (channels.${channelName}.input + "/nixos/modules"); };
+            { "${channelName}ModulesPath" = toString (channels.${channelName}.input + "/nixos/modules"); };
         in
         # Add `<channelName>ModulesPath`s
         (foldl' (lhs: rhs: lhs // rhs) { } (map f (attrNames channels)))
         # Override `modulesPath` because otherwise imports from there will not use patched nixpkgs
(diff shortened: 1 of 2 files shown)


blaggacao (approved):
Simple. Effective. Spot on. I would even say: neuralgic. :wink: What not.

An example to cater the _fully-featured_, maybe?

systemFlake.nix
```diff
@@ -149,7 +150,7 @@ let
             ];
           })
         ] ++ host.modules;
-        specialArgs = host.specialArgs;
+        specialArgs = host.specialArgs // { channel = channelInfo; };
       } // (optionalAttrs (host.output == "nixosConfigurations") {
         inherit lib baseModules specialArgs;
```
gytis-ivaskevicius, line 155: is this even working? Is it not getting overwritten by `host.output == "nixosConfigurations"` part? I might be misreading something.

Also, maybe it would make more sense to attach the name somewhere around here? https://github.com/gytis-ivaskevicius/flake-utils-plus/blob/6e6b93660f163e4fcb7f82f941c7733b3b200971/systemFlake.nix#L166-L173

gytis-ivaskevicius (approved):

systemFlake.nix
Pacman99: the specialArgs above has a `// host.specialArgs`. So theres no issue with specialArgs right now, but channel would get overwritten.

And yeah thats a good idea, I'll move `name` down.

blaggacao: I found that nuance a bit implicit. Glad to know we bump on it.

Pacman99: Ohh that doesn't actually work, since we can't pass a nixpkgs, I was just hoping to pass channel metadata, so its pulled directly from the `channels` argument.

systemFlake.nix
```diff
@@ -89,20 +89,24 @@ let
       selectedNixpkgs = getNixpkgs host;
       host = evalHostArgs (mergeAny hostDefaults host');
       patchedChannel = selectedNixpkgs.path;
+      channel = channels.${host.channelName} // { name = host.channelName; };
```
Pacman99, line 92: I just want to pass channel metadata and passing the imported channel would result in a string which is very hard to use. Whereas input/name is more useful. Theres no place right now to edit all channels, so I think this is the best way to do it.

systemFlake.nix
Pacman99: So I decided to go ahead and pass the instantiated nixpkgs and then added channel information - name and input - implicitly to all channels

PR #45, opened 2021-04-25
pass channels to overlays with three arguments

Pacman99:
just to add more flexibility to overlays, allow overlays to ask for `channels` as their first argument.

### DIFF
--- a/examples/fully-featured/flake.nix
+++ b/examples/fully-featured/flake.nix
@@ -86,6 +86,9 @@
         (final: prev: {
           neovim-nightly = neovim.defaultPackage.${prev.system};
         })
+        (channels: final: prev: {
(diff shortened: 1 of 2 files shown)


blaggacao:
I think this feature can deprecate the overlays builder. Since

Any function with a `channels: final: prev:` passed to channels.overlays = [];` now is effectively is at par in capabilities of the current overlaysBuilder.

blaggacao:
I think passing `pkgs` was done for the motive of accessing foreign channels, not foreign architectures (which wouldn't btw make a terrible lot of sense for building overlays).

Pacman99:
Closing this, I think its a bit to "magical" to include for a `overlaysBuilder` argument

PR #50, opened 2021-04-28
Option to not export overlay

Pacman99:
This adds the ability to create overlays that aren't exported. You just have to add an attribute called "__dontExport" and the overlaysFromChannelsExporter will filter it out.

As much as I like the auto-exporting overlays, it is kind of frustrating that there is no way out. The other solution is to try and pass the overlay to `inputs` which is needlessly complicated.
```
final: prev: {
  firefox = prev.firefox.override { privacySupport = true; };
  __dontExport = true;
}
```
vs
```
{ pkgs = self.pkgs; inputs = inputs // { internalOverlays.overlays.firefox = _: _: { firefox = null; }; }
```

### DIFF
--- a/examples/somewhat-realistic/flake.nix
+++ b/examples/somewhat-realistic/flake.nix
@@ -43,6 +43,13 @@
           # Overwrites specified packages to be used from unstable channel.
           inherit (channels.unstable) alacritty ranger;
         })
+        (final: prev: {
+          lib = prev.lib.extend (lfinal: lprev: {
+            utils = utils.lib;
+          });
+          # As long as the attribute exists, overlaysFromChannelsExporter won't export it
+          __dontExport = true;
+        })
         agenix.overlay
       ];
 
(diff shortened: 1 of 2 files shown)


gytis-ivaskevicius (commented):
I don't get it, what's the use case for this? From what I see it is just easier to just comment overlay out

Pacman99:
This is specifically for exporting overlays through overlaysFromChannelsExporter. Lets say I want to make an overlay to add a package, but I don't want to export that overlay. I can no longer make use of the fup exporting function, since that tries to export all overlays by default. So I added a "api" to the function that any overlays with the `__dontExport` property won't get exported.

Most of the time auto-exporting is good, you want to share all your creations. But there should be a way out for specific overlays.

blaggacao:
> But there should be a way out for specific overlays.

For example if you do overlays that back-port packages from an unstable channel, the level of ingenuity might just not be quite enough to advertise them as an overlay to others.

gytis-ivaskevicius (approved):
If you guys are fine with it - I am fine with it as well

PR #51, opened 2021-04-29
Rename channels to streams

nrdxp:
I realize this is a bit of a bikeshed, but words matter. After using Nix for 3 years it was enough for me to have an extreme reaction to the word "channels" and their associated updates. I don't think I'm alone in this. In fact, I'd be willing to bet that upstream explicitly removed any mention of the word from flakes for a very good reason. They wanted to make it clear that the painful impurities aligned with channels were a thing of the past.

I was a bit distraught to see that they are trying to be revived here, and so I thought I'd offer up a reasonable alternative. This may not matter much to newer Nix users, but to those who have used Nix stable for a little while, channels are the worst thing Nix has to offer. They break the contract of purity without even being clear about it, leaving you to discover that on your own after a few painful system upgrades.

With that said, I realize that the personal pronound is used extensively throughout this write up, and I understand it is an opinionated change that has 0 impact on usability. Therefore, I wouldn't blame you for rejecting my proposal outright. However, upstream is trying to kill the "channels" nomenclature for a good reason. Please don't keep it on life support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,8 +16,8 @@ This flake provides two main features (visible from `flake.nix`):
 - `nixosModules.saneFlakeDefaults` - Configures `nix.*` attributes. Generates `nix.nixPath`/`nix.registry` from flake `inputs`, sets `pkgs.nixUnstable` as the default also enables `ca-references` and `flakes`.
 - `lib.systemFlake { ... }` - Generates a system flake that may then be built.
 - `lib.exporter.modulesFromListExporter [ ./a.nix ./b.nix ]` - Generates modules attributes which looks like this `{ a = import ./a.nix; b = import ./b.nix; }`.
-- `lib.exporter.overlaysFromChannelsExporter channels` - Collects all overlays from channels and exports them as an appropriately namespaced attribute set. Users can instantiate with their nixpkgs version.
-- `lib.builder.packagesFromOverlayBuilderConstructor channels pkgs` - Similar to the overlay generator, but outputs them as packages, instead. Users can use your cache.
+- `lib.exporter.overlaysFromStreamsExporter streams` - Collects all overlays from streams and exports them as an appropriately namespaced attribute set. Users can instantiate with their nixpkgs version.
+- `lib.builder.packagesFromOverlayBuilderConstructor streams pkgs` - Similar to the overlay generator, but outputs them as packages, instead. Users can use your cache.
 
 
 # Examples #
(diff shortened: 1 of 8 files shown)


systemFlake.nix
```diff
@@ -5,8 +5,8 @@
 , supportedSystems ? flake-utils-plus.lib.defaultSystems
 , inputs
 
-, channels ? { }
-, channelsConfig ? { }
+, streams ? { }
+, streamsConfig ? { }
```
Pacman99, line 9:
```suggestion
, channels ? { }
, channelsConfig ? { }
, streams ? channels
, streamsConfig ? channelsConfig
```
For proper api deprecation protocol

Pacman99 (changes requested):
Target PR to staging.
I like the name, it doesn't have any reference to previous impurities in nix. It sounds really nice. And it still has the relation to "channels" so its self explanatory to anyone who has used channels. You hear the name and you think of nixpkgs versions.

flake.nix
```diff
@@ -9,7 +9,7 @@
       fupArgs = { flake-utils-plus = self; };
       systemFlake = import ./systemFlake.nix fupArgs;
       packagesFromOverlaysBuilderConstructor = import ./packagesFromOverlaysBuilderConstructor.nix fupArgs;
-      overlaysFromChannelsExporter = import ./overlaysFromChannelsExporter.nix fupArgs;
+      overlaysFromStreamsExporter = import ./overlaysFromStreamsExporter.nix fupArgs;
```
Pacman99, line 12: I don't think you renamed the file

Pacman99 (approved):
LGTM, just needs a file rename and `channels` deprecation lines

README.md
```diff
@@ -66,27 +66,27 @@ in flake-utils-plus.lib.systemFlake {
 
 
   ################
-  ### channels ###
+  ### streams ###
   ################
```
gytis-ivaskevicius, line 70: Misaligned lines 68 and 70 :D

systemFlake.nix
gytis-ivaskevicius: Not resolved. `channels` and `channelsConfig` will not work the way its defined now

systemFlake.nix
```diff
@@ -62,8 +64,8 @@ let
     "hosts"
     "hostDefaults"
     "nixosProfiles"
-    "channels"
-    "channelsConfig"
+    "streams"
+    "streamsConfig"
```
gytis-ivaskevicius, line 68: old `channels` and `channelsConfig` should not be removed for backward compatibility

gytis-ivaskevicius (changes requested):
What is the reason for choosing word `stream`?  Is this official? 

All in all, I am not a big fan of these changes - from my point of view it is adding more confusion to the end-user, word `stream` makes me think of Nats/Kafka/etc messaging and Rx pattern

Maybe something like `pkgsSet` would be more appropriate? (tho it sounds misleading as well :/ )

Anyways, I'd like to stick with the term `channels`, unless it was more or less officially renamed to something else or a better term pops up into somebody's mind

gytis-ivaskevicius:
Also actions arent running on your repo for some reason :thinking: https://github.com/nrdxp/flake-utils-plus/actions

README.md
nrdxp: > What is the reason for choosing word stream? Is this official?

Just some brainstorming. Also considered `nubs`, `pkgsets`, `stubs`, `sources`and `pkgstreams`. Decided since streams we're related to channels the association should be obvious enough.

As far as being official, well there is:
NixOS/nix#e2af11ce071fc109ad3e517bf15a87024d0a9ae5

Which explicitly removed channels from the manpages in favour of pulling from nixpkgs directly. I could have swore that, when flakes were brand new, there was some discussion about the move away from channels, but I can't seem to find it anywhere so it might be my imagination. It was about 2 years ago after all.

systemFlake.nix
nrdxp: forced pushed a fix, is that enough to do it?

gytis-ivaskevicius:
I gave some thought to this topic and decided that this changeset is not going to get merged. The whole point of fup is to simplify end-users system setup and creating custom terms just because it brings up bad memories is just counter-intuitive and overall in poor taste.

To quote @NobbZ
> Common nix literature still calls those branches we use as "channels", therefore I think it's counter productive to rename them in FU+
> Though indeed I'd appreciate if upstream/nix would use different nomenclature for the "declarative input branches for flakes", not really sure about the wording either

nrdxp:
For the record, with that last bit, NobbZ seems to agree with my poor taste :stuck_out_tongue_winking_eye:

In all seriousness, I understand it was a long shot anyway. Thanks for indulging me :+1:

PR #52, opened 2021-05-02
always use self.pkgs instead of channels

Pacman99:
this allows systemFlake to be called multiple times without having to pass `channels` each time. As long as it just uses self.pkgs, nixpkgs will only get evaluated once and all systemFlake calls use the same channels set

I'm thinking something like this:
```
let
  darwinSystems = systemFlake {
    inherit self inputs;
    hostDefaults = { builder = nix-darwin.lib.darwinSystem; };
    hosts = { Morty = { modules = [ ./Morty.nix; ]; }; };
  };
  nixosSystems = systemFlake {
    inherit self inputs;
    hosts = { Morty = { modules = [ ./Morty-nixos.nix ]; }; };
  };
in
systemFlake {
  inherit self inputs;
  inherit (darwinSystems) darwinConfigurations;
  inherit (nixosSystems) nixosConfigurations;
  channels = ...
}
```

### DIFF
--- a/systemFlake.nix
+++ b/systemFlake.nix
@@ -42,7 +42,7 @@ let
     { channelName ? "nixpkgs"
     , system ? "x86_64-linux"
     , output ? "nixosConfigurations"
-    , builder ? channels.${channelName}.input.lib.nixosSystem
+    , builder ? (getChannels system).${channelName}.input.lib.nixosSystem
     , modules ? [ ]
     , extraArgs ? { }
       # These are not part of the module system, so they can be used in `imports` lines without infinite recursion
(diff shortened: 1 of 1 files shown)


blaggacao:
Ui! Excellent motivation. I was wondering this for a while now, this looks like the cleanest solution.

blaggacao:
> As long as it just uses self.pkgs, nixpkgs will only get evaluated once and all systemFlake calls use the same channels set

I'm not an expert enough, but this looks on the surface like some really elegant fixed point magic, here. Correct?

Pacman99:
> I'm not an expert enough, but this looks on the surface of it like some really elegant fixed point magic, here. Correct?

Don't know a lot about fixed points in nix either, but it seems like thats the case. The fact that we can pass `self` to `mkFlake` then to `systemFlake` seems like fixed point magic.

gytis-ivaskevicius (approved):
Very simple and powerful. Bravo :)

PR #58, opened 2021-05-17
systemFlake: make one unified outputsBuilder

### LINKED ISSUE #56: `xyzBuilder = channels: {..}` should be replaced with `exports = channels: { packages = ...; apps = ...;}`
I have no idea why fup does not implement just a single builder for everything - but now that I look at it - it's obvious that this would be quite a bit nicer way of handling things. 

Possible issues:
Merging logic, it's always has been a pain. Tho if I recall correctly existing `mergeAny` function should just work out of the box

Expected outcome:
```nix
utils.lib.systemFlake {

  exports = channels: { 
    packages = { 
      inherit (channels.nixpkgs) hello coreutils neovim; 
    };
    
    apps = {
      neovim = mkApp {
        drv = channels.nixpkgs.neovim;
        exePath = "/bin/nvim";
      };   
    };

    checks = { 
      inherit (channels.nixpkgs) hello coreutils neovim; 
    };
    # ... etc ...
  };

}
```

@blaggacao @Pacman99 Looks good?

Pacman99:
replace output-specific builders

This is actually much simpler to implement than the old method. And the code looks waay better.

The hardest part is deprecating the current output-specific builders. I have no idea how to do that. I have tried many things like filtering for builders that return null. Either we end up with the flake just creating all the outputs, or infinite recursion when filtering.

implements/closes #56

### DIFF
--- a/src/systemFlake.nix
+++ b/src/systemFlake.nix
@@ -19,14 +19,21 @@
     extraArgs = sharedExtraArgs;
   }
 
-, outputsBuilder ? channels: { }
+, outputsBuilder ? channels: {
+  packages = packagesBuilder channels;
+  defaultPackage = defaultPackageBuilder channels;
+  apps = appsBuilder channels;
+  defaultApp = defaultAppBuilder channels;
+  devShell = devShellBuilder channels;
+  checks = checksBuilder channels;
+}
 
-, packagesBuilder ? null
-, defaultPackageBuilder ? null
-, appsBuilder ? null
-, defaultAppBuilder ? null
-, devShellBuilder ? null
-, checksBuilder ? null
+, packagesBuilder ? _: { }
+, defaultPackageBuilder ? _: { }
+, appsBuilder ? _: { }
+, defaultAppBuilder ? _: { }
+, devShellBuilder ? _: { }
+, checksBuilder ? _: { }
 , ...
 }@args:
 
--- a/src/systemFlake.nix
+++ b/src/systemFlake.nix
@@ -19,6 +19,8 @@
     extraArgs = sharedExtraArgs;
   }
 
+, outputsBuilder ? channels: { }
+
 , packagesBuilder ? null
 , defaultPackageBuilder ? null
 , appsBuilder ? null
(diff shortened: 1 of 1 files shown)


src/systemFlake.nix
```diff
@@ -19,12 +19,21 @@
...
-, devShellBuilder ? null
-, checksBuilder ? null
+, outputsBuilder ? channels: {
+  packages = packagesBuilder channels;
+  defaultPackage = defaultPackageBuilder channels;
+  apps = appsBuilder channels;
+  defaultApp = defaultAppBuilder channels;
+  devShell = devShellBuilder channels;
+  checks = checksBuilder channels;
+}
```
Pacman99, line 29: As you can imagine, this would have a couple consequences.
 - We create all the outputs whether users want them or not
 - I think `flake check` errors out from empty attrsets in defaultPackage and defaultApp

src/systemFlake.nix
```diff
@@ -189,21 +198,11 @@ mergeAny otherArguments (
...
-        mkOutput = output: builder:
-          mergeAny
-            # prevent override of nested outputs in otherArguments
-            (optionalAttrs (otherArguments ? ${output}.${system})
-              { ${output} = otherArguments.${output}.${system}; })
-            (optionalAttrs (args ? ${builder})
-              { ${output} = args.${builder} pkgs; });
+        outputs = mapAttrs
+          (outputName: output: otherArguments.${outputName}.${system} or { } // output)
+          (outputsBuilder pkgs);
```
Pacman99, line 203: I've tried filtering for empty attrsets or `null` here, but that can cause infinite recursion if any output depends on another.

src/systemFlake.nix
```diff
@@ -19,12 +19,21 @@
     extraArgs = sharedExtraArgs;
   }
 
-, packagesBuilder ? null
-, defaultPackageBuilder ? null
-, appsBuilder ? null
-, defaultAppBuilder ? null
-, devShellBuilder ? null
-, checksBuilder ? null
+, outputsBuilder ? channels: {
```
Pacman99, line 22: Naming: I really like `outputsBuilder`. I don't think we should use `exports`. Any better ideas?

src/systemFlake.nix
blaggacao: Maybe we can be a little more restrictive and require users to set `default*` in terms of `self.packages.${system}.mypackage` instead of provide the possibility of a full blown builder, that receives channels attribute. (Same for apps.)

What do you think?

src/systemFlake.nix
gytis-ivaskevicius: Maybe we can use something like this instead?
```
        deprecatedBuilders = channels: {}
          // optionalAttrs (packagesBuilder != null) (packagesBuilder channels)
          // optionalAttrs (defaultPackageBuilder != null) (defaultPackageBuilder channels)
          // optionalAttrs (appsBuilder != null) (appsBuilder channels)
          // optionalAttrs (defaultAppBuilder != null) (defaultAppBuilder channels)
          // optionalAttrs (devShellBuilder != null) (devShellBuilder channels)
          // optionalAttrs (checksBuilder != null) (checksBuilder channels);
```

src/systemFlake.nix
Pacman99: That is generally a good practice. I have no idea how to implement a restriction for that. If we try and check if something is referring to `self` we would probably run into infinite recursion.

gytis-ivaskevicius:
I don't want to ruin your fun but before bed, I thought to play around with it and I think I got a perfectly working version of this. I Will push in just a bit

Pacman99:
> I don't want to ruin your fun but before bed, I thought to play around with it and I think I got a perfectly working version of this. I Will push in just a bit

Please go ahead, if you want to open a separate PR thats fine too. I got something working, but there still some errors with my version.

gytis-ivaskevicius:
Yeah, I opened a separate one https://github.com/gytis-ivaskevicius/flake-utils-plus/pull/59
Would have edited this one but this PR is from your fork

I just need to double check CI output, but it should be good

PR #59, opened 2021-05-17
Add replace xyzBuilder with unified 'outputsBuilder'

gytis-ivaskevicius:


blaggacao (approved):

src/systemFlake.nix
```diff
@@ -189,24 +192,37 @@ mergeAny otherArguments (
...
+        mkOutput = output:
           mergeAny
             # prevent override of nested outputs in otherArguments
             (optionalAttrs (otherArguments ? ${output}.${system})
               { ${output} = otherArguments.${output}.${system}; })
-            (optionalAttrs (args ? ${builder})
-              { ${output} = args.${builder} pkgs; });
+            (optionalAttrs (systemOutputs ? ${output})
+              { ${output} = systemOutputs.${output}; });
+
```
Pacman99, line 213: I think this would be better done with a `mapAttrs`. That way its future proofed if more system-spaced outputs ever get added.

Pacman99: also code is a bit simpler, `mkOutput` is no longer necessary

gytis-ivaskevicius: Yes, I agree mapAttrs would be better option but we are dealing with multiple restrictions:
- `nix flake show` should not error - we cant output empty attrsets
- outputs must be merged - if we just go with mapAttrs while keeping in mind the first point - not all attrs will get merged.

And yeah, I guess I could go over the names and just generate attrset out of that. Probably will change that but I really should go and get some sleep now :D

Pacman99 (approved):

src/systemFlake.nix
Pacman99: sounds good, I think it could probably be used in the future once we remove the deprecated arguments. For now this works well.

gytis-ivaskevicius:
Also, not sure what happened to GitHub actions - but it does not show output for some reason :thinking: Anyways, there are two checks - one uses `xyzBuilder`, other `outputsBuilder` - they both result in exactly the same `nix flake show`

What needs to be tackled before merge:
- Add @Pacman99 suggestion
- Fix tests/examples to use `outputsBuilder`

If you guys wish to work on it - feel free to just push changes to this branch. But I'm pretty sure that ill find some time tomorrow to fix this up


On a totally unrelated note - @blaggacao thank you for talking me into adding devshell <3 It is so dam convenient

src/systemFlake.nix
gytis-ivaskevicius: Not going to do the attrs from list thingy. At least for time being - could not find a solution that I'd be happy with. Its other solutions are more complex and use around the same amount of lines

blaggacao:
Nice!

PR #65, opened 2021-05-21
modulesFromList: allow for import customization

Pacman99:
I have plans to try and export devshell modules, this is part of that plan.

### DIFF
--- a/src/modulesFromList.nix
+++ b/src/modulesFromList.nix
@@ -1,14 +1,17 @@
 { flake-utils-plus }:
 let
 
-  modulesFromListExporter = paths:
+  modulesFromListExporter = args:
     /**
       Synopsis: modulesFromListExporter _paths_
 
-      paths:    [ <path> <path> ]
+      paths:    [ <path> <path> ] or { paths = [ ]; _import = <function>; }
 
       Returns an attribute set of modules from a list of paths by converting
       the path's basename into the attribute key.
+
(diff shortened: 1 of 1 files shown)


src/modulesFromList.nix
```diff
@@ -23,6 +26,11 @@ let
 
     let
 
+      # To allow for the default to just pass a list
+      # or pass an attrset with `paths` and `_import`
+      paths = args.paths or args;
+      _import = args._import or import;
```
Pacman99, line 32: I did not know the `or` syntax works like this. If `args` is a list the `or` fallback still works.

gytis-ivaskevicius:
Looks good to me tho shall we add this while we are at it https://github.com/divnix/digga/blob/c162f5f46206346f431f3905642a8a9f17d42217/src/importers.nix#L3-L12

basically what I'm thinking of is function which accepts: 
- [ /path/to/file.nix ]
- { paths = [ /path/to/file.nix ]; }
- { paths = [ /path/to/file.nix ]; _import = customImporter; }
- { dir = /path/to/directory; }
- { dir = /path/to/directory; _import = customImporter; }

And probably assert is required so users could not define both paths and dir

gytis-ivaskevicius (approved):

Pacman99:
Is it ok to merge this for now as an incremental improvement?

I like your idea a lot, I will probably get around to implementing it soon, but it will make things a little more complicated.

Issue #62, opened 2021-05-20
Update repl.nix implementation


gytis-ivaskevicius:
Changes required:
- Make `nixosConfigurations` optional
- Turn it into a tiny overlay that is added by default. Maybe package name should be `fup-repl`, exec name `repl`? Any thoughts?
- CLI behavior:
1. `$ repl` - opens pretty much existing repl
2. `$ repl /path/to/flake.nix` - opens repl with specified flake loaded (I am really missing this feature :D )

More or less fixes https://github.com/divnix/digga/issues/17

blaggacao:
> Turn it into a tiny overlay that is added by default. Maybe package name should be fup-repl, exec name repl? Any thoughts?

I'm still trying to find a good reason for why it's not best option to keep it as a lib.

I think there are two broader use cases:

- Play the flake
- Play a specific host

For "play the flake", `self.pkgs` is probably fine to discover available packages.

For "play the host", I have no clear picture of what a user would want to do host-specific that only can be something with that particular config, right? I'm still repl-blind. Although, then `self.nixosConfigurations` would satisfy that need?

gytis-ivaskevicius:
I don't think I quite get your point. Mind elaborating on "play the host"? 
In case you are wondering why is it host aware - the explanation is simple: It is convenient to have a system `config` (parsed options) exported to inspect :)

blaggacao:
> In case you are wondering why is it host aware - the explanation is simple: It is convenient to have a system config (parsed options) exported to inspect :)

That was it! :+1: 

Would it make sense to provide a `configs` that holds all parsed `config` per host or something?

So to remain host aware, but make the repl itself only flake specific.

jared-w:
For reference, I borrowed heavily from this repo's `repl.nix` when designing one for a non nixos config project

```nix
let
  flake = builtins.getFlake (builtins.toString ../.);
  nixpkgs = flake.outputs.legacyPackages."${builtins.currentSystem}";
in {
  inherit flake;
} // flake // builtins // (builtins.removeAttrs nixpkgs [ "config" ])
// nixpkgs.lib
```

Essentially the difference is that it references the flake by path instead of self and doesn't overlay `flake.nixosConfigurations`. It's been quite helpful. I use it as `nix run .#repl` for this project where the app is shell script that runs the command `nix repl "$(git rev-parse --show-toplevel)"/nix/repl.nix`, but adding that shell script as a package is another alternative I suppose.

[...]

Issue #61, opened 2021-05-20
`channels.<name>.input` should be auto picked up if not explicitly defined


gytis-ivaskevicius:
Just a simplification for a very common use case. Should not be complicated to implement

https://github.com/gytis-ivaskevicius/flake-utils-plus/blob/staging/src/systemFlake.nix#L182-L193

Pacman99:
Ohh yeah we set this up in devos, should be simple enough to implement here.

Maybe we can add a `evalChannelArgs` to be able to set defaults for channel arguments, "channel submodule".

gytis-ivaskevicius:
@Pacman99 mind giving me some sort of example of this `evalChannelArgs`?

Pacman99:
```nix
evalChannelArgs = name:
  { input ? inputs.name
  , overlaysBuilder ? (_: [ ])
  , config ? { }
```
just a random idea, not sure if it should be done.

gytis-ivaskevicius:
Oh, I get it. But I don't think it brings much if any value? So I'm thinking that this should get implemented when some sort of use case comes up

gytis-ivaskevicius:
Resolved in staging

PR #66, opened 2021-05-23
Various fixes

### LINKED ISSUE #63: Get rid of saneFlakeDefaults
https://github.com/gytis-ivaskevicius/flake-utils-plus/blob/master/modules/saneFlakeDefaults.nix#L1-L18

Another little thing to simplify implementation for end users.
What needs to be done:
- Make sure that everything is easily overwritable
- Make sure that it does not conflict with non-nixos systems

Also, I notice that we have this logic duplication :thinking: 
https://github.com/gytis-ivaskevicius/flake-utils-plus/blob/staging/src/systemFlake.nix#L157
https://github.com/gytis-ivaskevicius/flake-utils-plus/blob/master/modules/saneFlakeDefaults.nix#L15

Related: 
https://github.com/gytis-ivaskevicius/flake-utils-plus/issues/60

gytis-ivaskevicius:
Resolves #63 and #61

### DIFF
--- a/lib/systemFlake.nix
+++ b/lib/systemFlake.nix
@@ -34,6 +34,16 @@ let
   inherit (flake-utils-plus.lib) eachSystem patchChannel mergeAny;
   inherit (builtins) foldl' mapAttrs removeAttrs attrValues attrNames listToAttrs concatMap;
 
+  fupOverlay = final: prev: {
+    fup-repl = final.writeShellScriptBin "repl" ''
+      if [ -z "$1" ]; then
+        nix repl ${./repl.nix}
+      else
+        nix repl --arg flakePath $(readlink -f $1 | sed 's|/flake.nix||') ${./repl.nix}
+      fi
+    '';
+  };
+
   filterAttrs = pred: set:
     listToAttrs (concatMap (name: let value = set.${name}; in if pred name value then [ ({ inherit name value; }) ] else [ ]) (attrNames set));
 
@@ -197,7 +207,7 @@ mergeAny otherArguments (
               __dontExport = true; # in case user uses overlaysFromChannelsExporter, doesn't hurt for others
               inherit srcs;
             })
-          ] ++ sharedOverlays ++ (if (value ? overlaysBuilder) then (value.overlaysBuilder pkgs) else [ ]);
+          ] ++ sharedOverlays ++ (if (value ? overlaysBuilder) then (value.overlaysBuilder pkgs) else [ ]) ++ [ fupOverlay ];
           config = channelsConfig // (value.config or { });
         }) // { inherit name; inherit (value) input; };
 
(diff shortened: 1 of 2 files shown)


gytis-ivaskevicius:
Also possibly closes https://github.com/gytis-ivaskevicius/flake-utils-plus/issues/62, not sure. Currently, repl is a little rough:
- lacks error handling
- lacks autocompletion

Also, could you guys play around with it, and check if you notice any bugs? Also I don't think that ill bother with issues above just yet, feel free to commit to this branch or create a seperate PR if interested :)

lib/systemFlake.nix
```diff
@@ -179,18 +190,24 @@ mergeAny otherArguments (
   eachSystem supportedSystems
     (system:
       let
+        filterAttrs = pred: set:
+          listToAttrs (concatMap (name: let value = set.${name}; in if pred name value then [ ({ inherit name value; }) ] else [ ]) (attrNames set));
+
+        channelFlakes = filterAttrs (_: value: value ? legacyPackages) inputs;
+        channelsFromFlakes = mapAttrs (name: input: { inherit input; }) channelFlakes;
```
gytis-ivaskevicius, line 197: Not related to this PR, just a thought:
```nix
        channelsFromInputs = _ inputs [
          (filterAttrs (_: value: value ? legacyPackages)) 
          (mapAttrs (name: input: { inherit input; }))
        ];
```
Do you guys think that function like this would get approved as part of nix builtins? I think in lib it is called `pipe` but it would be very nice to have it as builtin function which you don't need to import (like `derivation` or `toString`)

I picked function name `_` because it looks like a language feature and does not conflict  with anything

Pacman99 (approved):
I like it, its a really effective change set.

lib/systemFlake.nix
Pacman99: Yeah `pipe` is really useful, could also speed up evaluation if its done in nix.

blaggacao (commented):
This is not an immediate obvious change:

- The repl is delivered on the channel, even though in my line of thinking it is immediatly useful on it's own, for example in a devshell (and without any host deployed).

- It is not obvious how `channelsFromFlakes` feeds on `channels` instead of `inputs`, nor why.

---

I think I have to review a little more indepth in the coming days (I'm traveling, atm).

Pacman99:
> The repl is delivered on the channel, even though in my line of thinking it is immediatly useful on it's own, for example in a devshell (and without any host deployed).

repl is still accessible from lib. And since its just an overlay, you can access it in devshell. I think the package just has to be installed either in your hosts environment or in the devshell itself.

blaggacao:
In this case I guess, it would be sufficient to expose it as `self.overlays`, only.

We can't expose it as `self.packages`, since we don't depend on nixpkgs. Which leads me to another issue: we don't depend on anything that makes `nix repl` reproducible.

Then, users can enable the repl wherever they want through the overlay, and they are also responsible for meeting the implicit compatibility requirements (a compatible `nix` bin).

gytis-ivaskevicius:
Sounds good to me :)

gytis-ivaskevicius:
Merging. Additional changes will be done with separate PR's

PR #69, opened 2021-05-27
Add support for reverse DNS hostname declarations

blaggacao:
Implements naming `nixosConfigurations` like so:

`"com.example.myhost"` and automatically sets the `networking.domain` option if declared this way.

`nixosConfigurations."com.example.myhost"`

I chose reverese dns notation a) for proper grouping (e.g. in a repl) and b) since at this stage they alias a host's config, not precisely a host's IP, like normal DNS would.

### DIFF
--- a/lib/systemFlake.nix
+++ b/lib/systemFlake.nix
@@ -32,7 +32,24 @@
 
 let
   inherit (flake-utils-plus.lib) eachSystem patchChannel mergeAny;
-  inherit (builtins) foldl' mapAttrs removeAttrs attrValues attrNames listToAttrs concatMap;
+  inherit (builtins)
+    attrNames
+    attrValues
+    concatMap
+    concatStringsSep
+    elemAt
+    filter
+    foldl'
+    genList
+    head
+    isString
+    length
+    listToAttrs
+    mapAttrs
+    removeAttrs
+    split
+    tail
+    ;
 
   fupOverlay = final: prev: {
     fup-repl = final.writeShellScriptBin "repl" ''
@@ -48,10 +65,10 @@ let
     listToAttrs (concatMap (name: let value = set.${name}; in if pred name value then [ ({ inherit name value; }) ] else [ ]) (attrNames set));
 
   reverseList = xs:
-    let l = builtins.length xs; in builtins.genList (n: builtins.elemAt xs (l - n - 1)) l;
+    let l = length xs; in genList (n: elemAt xs (l - n - 1)) l;
 
   partitionString = sep: s:
-    builtins.filter (v: builtins.isString v) (builtins.split "${sep}" s);
+    filter (v: isString v) (split "${sep}" s);
 
 
   srcs = filterAttrs (_: value: !value ? outputs) inputs;
@@ -104,12 +121,12 @@ let
   configurationBuilder = reverseDnsFqdn: host': (
     let
       dnsLabels = reverseList (partitionString "\\." reverseDnsFqdn);
-      hostname = builtins.head dnsLabels;
+      hostname = head dnsLabels;
       domain = let
-        domainLabels = builtins.tail dnsLabels;
+        domainLabels = tail dnsLabels;
       in
         if domainLabels == [] then (lib.mkDefault null) # null is the networking.domain option's default
-        else builtins.concatStringsSep "." domainLabels;
+        else concatStringsSep "." domainLabels;
 
       selectedNixpkgs = getNixpkgs host;
       host = evalHostArgs (mergeAny hostDefaults host');
--- a/lib/systemFlake.nix
+++ b/lib/systemFlake.nix
@@ -47,6 +47,13 @@ let
   filterAttrs = pred: set:
     listToAttrs (concatMap (name: let value = set.${name}; in if pred name value then [ ({ inherit name value; }) ] else [ ]) (attrNames set));
 
+  reverseList = xs:
+    let l = builtins.length xs; in builtins.genList (n: builtins.elemAt xs (l - n - 1)) l;
+
+  partitionString = sep: s:
+    builtins.filter (v: builtins.isString v) (builtins.split "${sep}" s);
+
+
   srcs = filterAttrs (_: value: !value ? outputs) inputs;
 
   # set defaults and validate host arguments
@@ -94,8 +101,16 @@ let
   getChannels = system: self.pkgs.${system};
   getNixpkgs = host: (getChannels host.system).${host.channelName};
 
-  configurationBuilder = hostname: host': (
+  configurationBuilder = reverseDnsFqdn: host': (
     let
+      dnsLabels = reverseList (partitionString "\\." reverseDnsFqdn);
+      hostname = builtins.head dnsLabels;
+      domain = let
+        domainLabels = builtins.tail dnsLabels;
+      in
+        if domainLabels == [] then null # null is the networking.domain option's default
+        else builtins.concatStringsSep "." domainLabels;
+
       selectedNixpkgs = getNixpkgs host;
       host = evalHostArgs (mergeAny hostDefaults host');
       patchedChannel = selectedNixpkgs.path;
@@ -130,7 +145,7 @@ let
       }).config;
     in
     {
-      ${host.output}.${hostname} = host.builder ({
+      ${host.output}.${reverseDnsFqdn} = host.builder ({
         inherit (host) system;
         modules = [
           ({ pkgs, lib, options, config, ... }: {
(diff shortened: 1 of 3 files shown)


blaggacao:
> I don't think this kind of logic should be in systemFlake.

There reason for this to belong into `systemFlake` is the folowing:

1. Both, host label and domain are part of the same concern "host FQDN". 
2. FQDN here, in practice, is an alias for "host identity".
3. Previous to this PR, host identity could only be aliased via an incomplete host label part.
4. With this PR, host identity can be properly reflected in the `nixConfigurations` alias, due to the fact that reverse DNS and DNS are semantic equivalents.

Pacman99:
Thanks for the detailed explanation, I think I understand and mostly agree with your reasoning.

Just one concern, that I think you've already addressed in code, but I want to confirm:
If I'm a regular user that only has one or two hosts and I don't care about domains for my systems how will this change affect me?
What will `networking.domain` be if I just pass a regular hostname to fup? Can I change networking.domain, if I would prefer to set it manually?

lib/systemFlake.nix
```diff
@@ -141,6 +156,10 @@ let
                 networking.hostName = hostname;
               })
 
+              (optionalAttrs (options ? networking.domain) {
+                networking.domain = domain;
+              })
+
```
Pacman99, line 162:
```suggestion
                networking.hostName = lib.mkDefault hostname;
              })

              (optionalAttrs (options ? networking.domain) {
                networking.domain = lib.mkDefault domain;
              })

```
In case you don't want to pass domains to fup.

blaggacao: Good point! However, I guess the correct implementation I guess is slightly different, I'll push in a few seconds.

Pacman99: got it, yeah that looks right

blaggacao:
> If I'm a regular user that only has one or two hosts and I don't care about domains for my systems how will this change affect me?

Other than the latest fix, it wouldn't affect a regular user. While the string still is partitioned, it's only member will be the hostname label, so `builtins.head` recovers that.

And with the latest fix, domain is set to `lib.mkDefault null` mimicking the upstream default value.

lib/systemFlake.nix
```diff
@@ -94,8 +101,16 @@ let
   getChannels = system: self.pkgs.${system};
   getNixpkgs = host: (getChannels host.system).${host.channelName};
 
-  configurationBuilder = hostname: host': (
+  configurationBuilder = reverseDnsFqdn: host': (
```
Pacman99, line 104: In terms of naming, I thought `reverseDns` is generally the PTR record for IP to dns, but here it seems your using it in a more literal way as the dns string but flipped.

Can that be clarified with a comment or name change(flippedDns?)

Pacman99 (approved):
This makes a lot of sense to me now. It improves the situation for more advanced creators of hosts, while keeping it simpler for the rest of us.

Pacman99:
Is this ready to go? @gytis-ivaskevicius would be great if you could take a look.

lib/systemFlake.nix
```diff
@@ -94,8 +101,16 @@ let
...
     let
+      dnsLabels = reverseList (partitionString "\\." reverseDnsFqdn);
+      hostname = builtins.head dnsLabels;
+      domain = let
+        domainLabels = builtins.tail dnsLabels;
+      in
+        if domainLabels == [] then (lib.mkDefault null) # null is the networking.domain option's default
+        else builtins.concatStringsSep "." domainLabels;
```
gytis-ivaskevicius, line 112: Would be nice to move these builtins to the top like we do with the rest of them

lib/systemFlake.nix
```diff
@@ -94,8 +101,16 @@ let
   getChannels = system: self.pkgs.${system};
   getNixpkgs = host: (getChannels host.system).${host.channelName};
 
-  configurationBuilder = hostname: host': (
+  configurationBuilder = reverseDnsFqdn: host': (
     let
+      dnsLabels = reverseList (partitionString "\\." reverseDnsFqdn);
+      hostname = builtins.head dnsLabels;
+      domain = let
+        domainLabels = builtins.tail dnsLabels;
```
gytis-ivaskevicius, line 109: Makes me wonder if we actually need to reverse it, users could just type already reversed domain.

gytis-ivaskevicius: Tho its fine, if you guys prefer it we can definitely leave it as it is

gytis-ivaskevicius (approved):
Looks good :)
If you guys are in a hurry feel free to just merge it without those little tweaks

lib/systemFlake.nix
blaggacao: The reason to have reverse dns is two fold:

- It translates nicely and dorectly into a folder structure `./com/example/myhost`
- the domain grouping is preserved through alphanumerical sorting. For example in a repl exploring the `nixosConfigurations`

Pacman99: > It translates nicely and directly into a folder structure ./com/example/myhost (or any other nested attribute set)

Maybe we should reverse it in digga instead. For fup users it makes more sense for them to define their hosts with literal domain names.
```
hosts."myhost.example.com"
```
is much more obvious and simple to understand for in a fup flake.

> the domain grouping is preserved through alphanumerical sorting. For example in a repl exploring the nixosConfigurations

I don't see how this is exclusive to reverse dns.

lib/systemFlake.nix
Pacman99:
```suggestion
  configurationBuilder = fqdn: host': (
  # fqdn => example: myhost.example.com
```

lib/systemFlake.nix
blaggacao: > I don't see how this is exclusive to reverse dns. [useful alphanumerical sorting]

Semantically useful (grouped by zone):
```
com.example.a
com.example.b
de.example.a
de.example.b
```

Not very useful:
```
a.example.com
a.example.de
b.example.com
b.example.de
```

gytis-ivaskevicius: One may argue that it would be more convenient this way due to people straight up typing the machine name is more unique compared to the domain
```
GytisMachine.xyz.example.com
PacmanMachine.xyz.example.com
DavidMachine.xyz.example.com
```

blaggacao: I'm still trying to figure out if I'm only upholding a "theoretical" benefit here.

My gut tells me there is something valid to reverse DNS, but sometimes my gut is wrong.

I'll meditate over it. :wink:

One thing that might be bothersome is that the api between `digga` and `fup` might diverge. Although maybe the magic digga importer can already deal with it.

Still meditating.

Pacman99: > One thing that might be bothersome is that the api between digga and fup might diverge. Although maybe the magic digga importer can already deal with it.

This is only true for the auto-importer where it makes sense for folder structure to use reverse dns - it would be awkward to have a `com.nix` in a folder.

I do see some historical use of reverse dns in programming, freedesktop and java. It seems to be pretty common in programming, so perhaps we should just stick with it. Those groups seem to have identified advantages with reverse dns notation. But it does make fup user facing api kind of complicated.

blaggacao: So I tossed a coin and it said: reverse DNS. So that it'll be. :wink:

blaggacao:
bors try

lib/systemFlake.nix
blaggacao: Wiki has it as: https://en.wikipedia.org/wiki/Reverse_domain_name_notation

lib/systemFlake.nix
Pacman99: cool, can you at least add a comment under the function argument to clarify what you mean by reverse dns, like I suggested in the other comment(adapted to reverse dns of course).
Reverse dns is too ambigous and confusing of a name.

lib/systemFlake.nix
Pacman99: Maybe reverseFqdn or reverseFqdnNotation
```suggestion
  configurationBuilder = reverseFqdn: host': (
  # example: com.example.myhost
```

blaggacao: You are right, addressed in https://github.com/gytis-ivaskevicius/flake-utils-plus/pull/69/commits/e834947e8b2bacee6fe2880fbfa557b5ef539ff8

lib/systemFlake.nix
blaggacao: Addressed in https://github.com/gytis-ivaskevicius/flake-utils-plus/pull/69/commits/0d915599218f910c63ced0c3d0c1732f72b98c30

PR #70, opened 2021-06-01
fix: don't append overlays manually when passing pkgs

Pacman99:
overlays get appended by nixpkgs module
creates problems with overlays that apply patches, since they try to get applied twice.

### DIFF
--- a/lib/systemFlake.nix
+++ b/lib/systemFlake.nix
@@ -190,7 +190,7 @@ let
                       import patchedChannel
                         {
                           inherit (host) system;
-                          overlays = selectedNixpkgs.overlays ++ hostConfig.nixpkgs.overlays;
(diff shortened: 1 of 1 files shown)


Pacman99:
A little more context: https://github.com/NixOS/nixpkgs/blob/9b4ee111e66b406156755e7be65ac306d3ddade3/nixos/modules/misc/nixpkgs.nix#L62

chvp:
This is correct, the `hostConfig.nixpkgs.overlays` was added [here](https://github.com/gytis-ivaskevicius/flake-utils-plus/pull/54). [This commit](https://github.com/gytis-ivaskevicius/flake-utils-plus/commit/a79e19f36dd428ccfd2415ed14801b15eb2f520a) to be precise. I noticed the same issue when I made my changes.

PR #72, opened 2021-06-15
da/learn names from module meta

blaggacao:
A laid back import has a disadvantage:

Downstream users that choose to import at the code entrypoint have to carry those like a callback all the way trough the code-path to `fup`. So no unified import at that entry-point is possible.

With this change, users can design code that imports any module as soon as they wish as long as it honors the ubiquitous `_file` key convention.

- Collateral downstream benefit of construing compatible modules: error reporting becomes more user friendly (no `<unnokwn-file>` anymore).

---

- Revert "modulesFromList: allow for import customization"
- modulesFromList: infer names from paths or modules's _file attr

### DIFF
--- a/lib/modulesFromList.nix
+++ b/lib/modulesFromList.nix
@@ -1,17 +1,14 @@
 { flake-utils-plus }:
 let
 
-  modulesFromListExporter = args:
+  modulesFromListExporter = paths:
     /**
       Synopsis: modulesFromListExporter _paths_
 
-      paths:    [ <path> <path> ] or { paths = [ ]; _import = <function>; }
+      paths:    [ <path> <path> ]
 
       Returns an attribute set of modules from a list of paths by converting
       the path's basename into the attribute key.
-
-      Optionally, an attrset can be passed containing `paths` and `_import`
-      to control how paths get imported
 
       Example:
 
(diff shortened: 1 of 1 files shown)


lib/modulesFromList.nix
```diff
@@ -43,14 +35,40 @@ let
...
-        value = _import path;
-      })
-      paths;
+      (arg:
+
+        # a regular path to be imported
+        if builtins.isPath arg then
+          {
+            name = removeSuffix ".nix" (baseNameOf arg);
+            value = import arg;
```
gytis-ivaskevicius, line 50: I am not a big fan of the removal of `_import`. Was hoping to add recImportAttrs function here as well :/

gytis-ivaskevicius (approved):
Looks good to me :+1: I guess `_import` can be fixed sometime in the future

lib/modulesFromList.nix
Pacman99: I consider this function to be more of an "exporter" if that makes sense. I think a recImport is an _importer_ and should be separate. We did an implementation for that in digga with rakeLeaves, maybe we can extract that into fup im just not sure where it would be useful for end users outside the digga context.

lib/modulesFromList.nix
```diff
@@ -43,14 +35,40 @@ let
...
+          }
+
+        # a module function with a _file attr
+        elseif ((builtins.isFunction arg) && (hasFileAttr (peek args)))
```
Pacman99, line 54: Didnt know you could do else if like this, I thought they had to seperated in nix.

Pacman99 (approved):
Cool stuff I like the goal

PR #73, opened 2021-06-20
#62

gytis-ivaskevicius:
Resolves https://github.com/gytis-ivaskevicius/flake-utils-plus/issues/62

### DIFF
--- a/lib/overlay.nix
+++ b/lib/overlay.nix
@@ -1,0 +1,25 @@
+final: prev:
+let
+  example = command: desc: ''\n\u001b[33m ${command}\u001b[0m - ${desc}'';
+in
+{
+  __dontExport = true;
+
+  fup-repl = final.writeShellScriptBin "repl" ''
+    case "$1" in
+      "-h"|"--help"|"help")
+        printf "%b\n\e[4mUsage\e[0m: \
+          ${example "repl" "Loads system flake if available."} \
+          ${example "repl /path/to/flake.nix" "Loads specified flake."}\n"
+      ;;
+      *)
+        if [ -z "$1" ]; then
+          nix repl ${./repl.nix}
+        else
+          nix repl --arg flakePath $(readlink -f $1 | sed 's|/flake.nix||') ${./repl.nix}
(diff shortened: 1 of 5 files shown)


lib/systemFlake.nix
```diff
@@ -244,7 +234,7 @@ mergeAny otherArguments (
               __dontExport = true; # in case user uses overlaysFromChannelsExporter, doesn't hurt for others
               inherit srcs;
             })
-          ] ++ sharedOverlays ++ (if (value ? overlaysBuilder) then (value.overlaysBuilder pkgs) else [ ]) ++ [ fupOverlay ];
+          ] ++ sharedOverlays ++ (if (value ? overlaysBuilder) then (value.overlaysBuilder pkgs) else [ ]) ++ [ flake-utils-plus.overlay ];
```
blaggacao, line 237: I have a sense that it would be better to let the user include `flake-utils-plus.overlay`.

The overlays and tools might eventually grow over time, so to preserve the minimal character of `fup` I opine this should rather not be added by default.

Ignore me, if you don't agree.  :smile:

lib/overlay.nix
```diff
@@ -0,0 +1,25 @@
+final: prev:
```
blaggacao, line 1:
```suggestion
self: final: prev:
```

In the form of a proto-overlay, I beleive this would obsolete the need for a contrived indirection through the registry.

Exposed as `self.replOverlay` to avoid overlay checks.

blaggacao (commented):
Acknowledging that my picture is far from complete, here are my thoughts.

(Informed mostly by her majesty: my gut)

:smile:

lib/systemFlake.nix
blaggacao: `(flake-utils-plus.replOverlay self)` &mdash; according to the other comment.

gytis-ivaskevicius: The package name will not overwrite anything and I'd really rather leave as little as possible to the user to do thus this changeset stays

Issue #77, opened 2021-06-30
The default for `experimental-features` is not fully enabling CA


NobbZ:
The current default for `experimental-features` enables `ca-derivations`, which enables one to **build** content addressed derivations, but not to have them as inputs to other derivations, regardless of them beeing CA or IA.

Therefore please also enable `ca-references` by default.

blaggacao:
> Therefore please also enable ca-references by default.

I'd rather argue for removing anything non-flake related. Other feature flags are orthogonal to this entire repo and people would be rightfully surprised to find them set here.

NobbZ:
As CA was enabled, I assumed, its actually required for FU+. If this is not the case, I am the same opinion as David, please remove it completely and make it easier for users to add experimental features as suggested in #76.

PR #78, opened 2021-07-03
Add mkDefault to extra-options

### LINKED ISSUE #76: Make it easier to override "`experimental-features`"
Currently the `experimental-features` setting for nix is hardcoded to a specific set of items.

https://github.com/gytis-ivaskevicius/flake-utils-plus/blob/staging/lib/systemFlake.nix#L198-L200

This makes it hard to change the values, eg. to add `recursive-nix` or `ca-references` to the list, or any other not yet known experimental feature.

Therefore, it might be sensible to change the current implementation in one of these ways:

1. Use `mkDefault` instead, then a user can easily overwrite the `extraOptions` for nix and does not have to use `mkForce`, which again might even break other stuff, as eg. the `hydra` options wouldn't be able anymore to add their required settings, while at the same time , enabling hydra would disable any `mkDefault` enabled stuff.
2. Provide an option, `nix.experimentalFeatures`, which can be set instead, and merges from individual strings to "string concatenated with spaces" (`types.separatedString " "`). This should be more compatible with other modules that touch `nix.extraOptions`.

gytis-ivaskevicius:
Resolve #76 #77
I left `ca-references` in since I am not sure why it is there and I am afraid to remove it. 
If I recall correctly at some point `ca-references` were required for flakes for some reason thus was erroring out for me. Maybe it was a bug or some odd config. Not sure.

### DIFF
--- a/lib/repl.nix
+++ b/lib/repl.nix
@@ -29,4 +29,4 @@ in
 // (flake.nixosConfigurations or { })
 // flake.nixosConfigurations.${hostname} or { }
 // nixpkgsOutput
-  // { loadFlake = path: getFlake (toString path);}
+  // { loadFlake = path: getFlake (toString path); }
--- a/lib/systemFlake.nix
+++ b/lib/systemFlake.nix
@@ -196,7 +196,7 @@ let
               })
 
               (optionalAttrs (options ? nix.extraOptions) {
-                nix.extraOptions = lib.mkDefault "experimental-features = nix-command ca-references flakes";
+                nix.extraOptions = "extra-experimental-features = nix-command ca-references flakes";
               })
 
               {
--- a/lib/systemFlake.nix
+++ b/lib/systemFlake.nix
@@ -196,7 +196,7 @@ let
               })
 
               (optionalAttrs (options ? nix.extraOptions) {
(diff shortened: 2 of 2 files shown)


Pacman99 (approved):

gytis-ivaskevicius:
There is a little issue with this - if somebody defines `nix.extraOptions` - no matter what it has it overwrites this option. Some fancy merge function should be defined so it would not apply this option if `experimental-features` is already defined

blaggacao:
> `ca-references`

I think this can be removed if we don't have proof-backed insight why it's necessary for flakes. As far as I understand it `ca-references` have their own set of issues.

> `experimental-features`

-> `extra-experimental-features` &mdash; works at least on the command line as expected

blaggacao (approved):

fufexan (approved):

gytis-ivaskevicius:
I took @blaggacao advice and implemented it using `extra-experimental-features` option. 
Not removing ca-references since this issue is still open https://github.com/NixOS/nix/issues/3422

Not so fun fact:
CI seems to be broken, it is not reporting errors :/

blaggacao:
w.r.t. `ca-references` it didn't work without it on my machine, either. Thanks for the pointer!

Issue #79, opened 2021-07-09
Home Manager support


fufexan:
I'm thinking of writing something similar to `hosts`, but for standalone Home Manager. I believe it's not just me who would like to see this go through, as the difference between configuring `nixosConfigurations` and `homeManagerConfigurations` is a bit stark at the moment.
Would this benefit the project or is it something better suited in the user's personal `lib`?

blaggacao:
I'd welcome this addition. It's fully prototyped on `divnix/digga`, already.

blaggacao:
In fact, we are currently figuring out how to properly use `useUserPackages`. So far my conclusion is:

- `user@host` type profiles should probably use the same config as the host and should only be used for the host they where made.

- a portable `user` should use `useUserPackages = false` that is using `$HOME/.nix-profile` to remain portable for the "my user config at work" use case.

The first can be easily derived from a host's `config.home-manage.users`. The latter is a little more involved and I'm trying to figure out the best design, here: it would be a list of users that would be built from a certain lists of modules and profiles.

jessestricker:
@gytis-ivaskevicius Is this something you like to see included in FUP? If so, I would be willing to give implementing this a try. 

Or would a feature like this be too much "fluff"? 🤔

gytis-ivaskevicius:
Yeah, I'd be down for this

PR #82, opened 2021-07-12
imp: make system output dynamic

blaggacao:
towards #79

### DIFF
--- a/devShell.nix
+++ b/devShell.nix
@@ -27,8 +27,8 @@ let
       set -e
       cd $DEVSHELL_ROOT/tests/${name}
       nix flake lock --update-input utils
-      nix flake show
-      nix flake check
+      nix flake show "$@"
+      nix flake check "$@"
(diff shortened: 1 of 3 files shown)


gytis-ivaskevicius:
Not suited for HM. Does not exactly solve anything since it lacks arguments that should be used to call builder :/

Maybe we could just code it dynamically so all `outputBuilder`  values would get passed to the main flake? Should not be terribly difficult

gytis-ivaskevicius (approved):
Looks great! But did you execute tests just to be sure?

blaggacao:
> Looks great! But did you execute tests? (Just making sure)

I was doing that right now and this seems to trigger an infinite recursion. I'll report back as I learn more.

Pacman99 (approved):
nice! Yeah this works well

blaggacao:
Both, repo tests as well as downstream tests succeeded with the latest fixup.

PR #84, opened 2021-07-17
Restructure exporters

fufexan:
* move common functions to `lib/internal-functions.nix`
* rename
```hs
exporters.modulesFromList  -> exportModules
exporters.fromOverlays     -> exportPackages
exporters.internalOverlays -> exportOverlays
systemFlake                -> mkFlake
```
* `exportModules` now supports recursively exporting directories, thanks to [digga's `rakeLeaves`](https://github.com/divnix/digga/blob/develop/src/importers.nix#L61)


fufexan:
what the fuck did I do lol

gytis-ivaskevicius (commented):
Looks good to me, but I'd like to get some input from devos guys on this

flake.nix
```diff
@@ -22,10 +29,10 @@
 
       devShell.x86_64-linux = import ./devShell.nix { system = "x86_64-linux"; };
 
-      lib = flake-utils.lib // {
-        # modulesFromList is deprecated, will be removed in future releases
-        inherit systemFlake modulesFromList;
+      lib = flake-utils.lib // internal-functions // {
```
blaggacao, line 32: I would advise against exporting all of `internal-functions`: it puts unnecessary extra burden on backwards compatibility and deprecation requirements. It's a well-known truth that people inevitable start to depend on public APIs, so let's be careful about it.

I would only export the main functions of this library. Since `rakeLEaves` is now included here (:+1:), I'd also export this specifically, since we make heavy use in devos flakes.

Another function that's useful outside of `fup` is the one that reverses a hostname into a reverse dns notation. I already found myself re-implementing this in various versions downstream.

blaggacao (approved):
Just a few things for consideration. All tests pass? Can't wait to get this in before we merge `digga/develop` back in `digga/main`.

flake.nix
gytis-ivaskevicius: These internal functions can either be in namespace and be like "if you are using it - your problem" or each file can import it manually. Or at least these are solutions I see

fufexan: I will export the main functions in the top-level lib, and if you're alright with it, add `lib.internal` which contains everything inside.

fufexan: By the way @blaggacao how do you suggest I add the `reverseDns` function? I feel like it's a bit scattered in `mkFlake`. Should I combine `domain`, `dnsLabels` and `domain`?

Maybe there's some more to the functionality which I haven't discovered yet.

gytis-ivaskevicius:
Can you tweak workflow or push this branch to this repo to see CI result? After that I guess we can merge?

flake.nix
blaggacao: @fufexan Now, that I look at it again, maybe we should leave out the reverse dns thing. It's already too bound to the nixos networking module.

fufexan:
@blaggacao what do you say, should I rename
```nix
packagesFromOverlaysBuilder -> exportOverlaysBuilder
packagesFromOverlays        -> exportOverlays
```
in `lib/exportOverlays.nix`?

blaggacao:
If any, the whole thing should be named `exportPackages`, but yes, that would be fine.

fufexan:
Then what about the already-existing `lib/exportPackages`?

flake.nix
```diff
@@ -23,12 +30,17 @@
       devShell.x86_64-linux = import ./devShell.nix { system = "x86_64-linux"; };
 
       lib = flake-utils.lib // {
-        # modulesFromList is deprecated, will be removed in future releases
-        inherit systemFlake modulesFromList;
+        inherit (internal-functions) rakeLeaves;
```
gytis-ivaskevicius, line 33: I don't think it's supposed to be here?

gytis-ivaskevicius (approved):

flake.nix
fufexan: Why not? Where should it be then?

I thought we were removing deprecations in a later PR.

gytis-ivaskevicius: I mean the `rakeLeaves` function. I do not see a reason why it should not be internal only since modules exporter takes care of use-cases involving this function

fufexan: @blaggacao said they use it a lot at DevOS and digga. So it's better to have it top-level.

gytis-ivaskevicius: I got a feeling that they themselves would prefer using something along the lines of the proxy pattern. I'd say there is no reason to export it externally, thus lets avoid that. Especially since we already have an `internal` namespace

fufexan: @blaggacao thoughts?

blaggacao: I'm fine with proxying via the `internal` namespace. These projects are so close, that we probably would not have issue with drift.

fufexan: Okay, fixing now.

blaggacao (approved):
LGTM

lib/exportModules.nix
```diff
@@ -21,35 +30,20 @@ let
...
-
-    in
-
     genAttrs'
       (arg:
 
-        # a regular path to be imported
-        if builtins.isPath arg then
+        # a regular file to be imported
+        if pathIsRegularFile arg then
```
blaggacao, line 37: Unfotunately the implementatino does not work with `arg` being a (module) function, which is perfectly legal, as lon as it exhibits a `_file` attribute.

I'll push a fix.

lib/exportModules.nix
```diff
@@ -21,35 +30,20 @@ let
...
           {
             name = removeSuffix ".nix" (baseNameOf arg);
             value = import arg;
           }
 
+        # a directory to be recursively imported
+        else if pathIsDirectory arg then
+          rakeLeaves arg
```
blaggacao, line 45: Just noticed: since this is intended to be consumed by `self.nixosModules`, this also needs to implement `digga.flattenTree`. Otherwise this will produce illegal (nested) output. See `flake check`.

blaggacao: unfortunately, this seems to cause a regression:
```console
error: 'name' attribute missing in a call to 'listToAttrs'

       at /nix/store/qqv2yvbj4ra9hpf8ar4bih9wc1fxwvs0-source/lib/internal-functions.nix:13:29:

           12|   */
           13|   genAttrs' = func: values: builtins.listToAttrs (map func values);
             |                             ^
           14|

       … while evaluating 'genAttrs''

       at /nix/store/qqv2yvbj4ra9hpf8ar4bih9wc1fxwvs0-source/lib/internal-functions.nix:13:21:

           12|   */
           13|   genAttrs' = func: values: builtins.listToAttrs (map func values);
             |                     ^
           14|

       … from call site

       at /nix/store/qqv2yvbj4ra9hpf8ar4bih9wc1fxwvs0-source/lib/exportModules.nix:33:5:

           32|
           33|     genAttrs'
             |     ^
           34|       (arg:

       … while evaluating 'exportModules'

       at /nix/store/qqv2yvbj4ra9hpf8ar4bih9wc1fxwvs0-source/lib/exportModules.nix:13:19:

           12|     ;
           13|   exportModules = args:
             |                   ^
           14|     /**

       … from call site

       at /nix/store/c2jawjpd85s319ql9n4fsnc88krbbjsj-source/src/mkFlake/fup-adapter.nix:88:23:

           87|
           88|     devshellModules = flake-utils-plus.lib.exportModules config.devshell.modules;
             |                       ^
           89|

       … while checking flake output 'devshellModules'

       at /nix/store/c2jawjpd85s319ql9n4fsnc88krbbjsj-source/src/mkFlake/fup-adapter.nix:88:5:

           87|
           88|     devshellModules = flake-utils-plus.lib.exportModules config.devshell.modules;
             |     ^
           89|
```

I'll be reverting. Maybe `rakeLeaves` is not the right tool, here.

PR #89, opened 2021-09-05
Revert auto-channel detection that produced irrecoverable edge-cases

blaggacao:
In more advanced usage scenarios, the guarantees of this code path
are never strong enough and the error produced is non-recoverable.

Sorry to lovers of "automagic".

### DIFF
--- a/examples/exporters/flake.nix
+++ b/examples/exporters/flake.nix
@@ -16,6 +16,7 @@
       inherit self inputs;
 
       # Channel specific overlays. Overlays `coreutils` from `unstable` channel.
+      channels.unstable = {};
       channels.nixpkgs.overlaysBuilder = channels: [
         (final: prev: { inherit (channels.unstable) ranger; })
       ];
--- a/examples/home-manager+nur+neovim/flake.nix
+++ b/examples/home-manager+nur+neovim/flake.nix
@@ -26,6 +26,7 @@
 
 
       channelsConfig.allowUnfree = true;
+      channels.nixpkgs = {};
 
       sharedOverlays = [
         nur.overlay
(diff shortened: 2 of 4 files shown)


Pacman99:
Can you provide an example of such an irrecoverable error, to strengthen the argument for this change.

blaggacao:
> Can you provide an example of such an irrecoverable error, to strengthen the argument for this change.

`filterAttrs (_: value: value ? legacyPackages && value.legacyPackages.x86_64-linux ? nix) inputs;`

any item that unintentionally falls thogh that predicate will suffice.

With the current implementation, a repo that bears `legacyPackages.x86_64-linux.nix`.

But pushing this formally incorrect code further to the edges, just modifies the predicate for the failure scenario.

I'm in a situation a.t.m. for my own "safety" where I can't accept lingering regressions.

blaggacao:
Here is another good one:

The expresivity to be able to _canonically_ encode the intent to _not_ use a nixpkgs type input _also_ as a channel must be preserved for a low level lib.

Use case: don't encourage you users to use a channel you never intended to expose.

Pacman99 (approved):
I'm fine with this, I see the edge case problems. I figure that it would just be better to not infer and just do `inputs.${channelName}` so that would just trigger a nix error saying that the attribute doesn't exist in inputs immediately guiding the user to the problem. I think the current inference methods are cool but extra.

I also personally don't think this kind of magic should be here anyways, but I'm alright with it if done right.

mstrangfeld:
Just wanted to add that this currently happens when having the current version of [agenix](https://github.com/ryantm/agenix) in the inputs. Was pretty hard to figure out what was actually causing it... 😅 
```
error: attribute 'x86_64-linux' missing

       at /nix/store/i5ipw8rvj1fvdaj01bqzc0dpw9irrb4h-source/lib/mkFlake.nix:208:74:

          207|         # For some odd reason `devshell` contains `legacyPackages` out put as well
          208|         channelFlakes = filterAttrs (_: value: value ? legacyPackages && value.legacyPackages.x86_64-linux ? nix) inputs;
             |                                                                          ^
          209|         channelsFromFlakes = mapAttrs (name: input: { inherit input; }) channelFlakes;

```

PR #90, opened 2021-09-05
fix config handling revert complex patch handling

blaggacao:
Revert patching functionality and fix config handling

The patching functionality is fundamentally ill designed
and makes the code complex to the point that it is increasingly
hard to audit it for correctness.

The root of evil and ill-design is that it is a work-around for
what the Unoficial Flake Roadmap describes as follows:
 - flake inputs are flake inputs
 - patched flake inputs are flake inputs
 
 Furthermore this commit fixes a naive implementation of nixpkgs.config
 handling that causes nasty regressions on edge cases. Namely, not only
 can a nixpkgs config be a function, but also the merging semantics
 as defined by the module system requires detailed knowledge of the
 potential shape of the cofig attrs / function.


Pacman99:
I agree with this, but I realize a fair bit of contributors to here really value the patches feature. I think we need some discussion on patching and what people use it for.

shism2 (approved):

Issue #98, opened 2021-10-25
`overlayNames` errors; can't necessarily pass `null` to an overlay


Radvendii:
`overlayNames` calls the `overlay` attribute with `null null` for the arguments `final` and `prev`. This works if the `overlay` attribute looks like:

```nix
{
  overlay = final: prev: {
    blah = prev.callPackage ./blah {};
    foo = prev.callPackage ./foo {};
  };
}
```
because nix is lazy, so it will not even evaluate the attribute values.

However, if the overlay looks something like:

```nix
{
  overlay = final: prev:
    prev.lib.genAttrs 
      (builtins.readDir ./packages) 
      (pkgName: _: prev.callPackage ./packages + "/${pkgName}");
}
```

Then the overlay needs access to `prev.lib` *before* the set of names is known, and so `overlayNames` errors with `value is null while a set was expected`

For reference, this is where `overlayNames` is defined: https://github.com/gytis-ivaskevicius/flake-utils-plus/blob/master/lib/exportOverlays.nix#L63

Radvendii:
Yeah, that seems good. I should say that I wasn't using this directly, I was using github.com/divnix/digga, which called the function. So maybe this problem should be resolved there.

I guess it should at least be documented that this cannot be called on an arbitrary overlay.

PR #100, opened 2021-11-06
Issue/98

gytis-ivaskevicius:
Should improve #98. 
Added a couple of extra checks which hopefully will catch at least a few edge cases. Unless somebody else has any suggestions - there is nothing else one may do for that particular issue.

Also for the record, I have not tested that function much, maybe someone could confirm that it works as expected?


Pacman99 (approved):
Yeah this is a pretty effective solution for this, thanks!
But I'm really not a huge fan of the complexity of this solution, and the growing complexity of exportOverlays, which I know is largely on me.
I'm starting to feel like automatic exporting isn't the best idea. I quite like how clean the selective exporting of modules in digga has started to become.

lib/exportOverlays.nix
```diff
@@ -44,27 +44,44 @@ let
...
         foldl'
         head
+        isAttrs
+        isFunction
         listToAttrs
         mapAttrs
+        tryEval
         ;
+
+      # Hopefully will fix a couple of edge cases. Even tho `lib` is defined as API here - it is not. Do not use it.
```
Pacman99, line 59: Is this note meant for developers of exportOverlays? or users of `fakePkgs`

Pacman99:
This only somewhat deals with the issue, the larger culprit will likely be `exportPackages`

lib/exportOverlays.nix
gytis-ivaskevicius: I guess users if they decide to check the source

PR #110, opened 2022-01-14
Add function, genPkgOverlay, to create an overlay for an input's package

Pacman99:
This is a function to make it easier to add packages from flakes that don't have an `overlay` output. And its also useful if you prefer using the the exported package from a flake instead of the overlay. There are some good reasons for doing this: cache hits, reproducibility, overlay is failing with your channel.

I was thinking of making this an api argument for fup or digga, but I think a lib function is the cleanest way of doing this. I can't think of a good name, so suggestions are welcome.

### DIFF
--- a/examples/home-manager+nur+neovim/flake.nix
+++ b/examples/home-manager+nur+neovim/flake.nix
@@ -10,7 +10,6 @@
 
     neovim = {
       url = github:neovim/neovim?dir=contrib;
-      inputs.nixpkgs.follows = "nixpkgs";
     };
 
     home-manager = {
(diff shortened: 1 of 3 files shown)


gytis-ivaskevicius:
huh, simple enough :)

gytis-ivaskevicius (approved):

gytis-ivaskevicius:
Actually, should we add this? Maybe this is a bit of bloat since we are talking about simple `inherit` syntax

Pacman99:
You actually can't use inherit for this, because the ${} syntax isn't allowed inside the parenthesis.
I'm open to other ideas. The main goal is to make it easy to add packages from flakes when fup is designed around overlays. I just feel like it looks really ugly to create an overlay so I was thinking it would be nice to have a function generate the overlay.

PR #111, opened 2022-01-31
Export `autoGenFromInputs` module for `nix-darwin`

Enzime:
I recently realized that there's little difference between NixOS,
nix-darwin and home-manager's module systems, so modules should work
across them as long as the options are implemented.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -19,6 +19,7 @@
       inherit overlay;
 
       nixosModules.autoGenFromInputs = import ./lib/options.nix;
+      darwinModules.autoGenFromInputs = import ./lib/options.nix;
 
       devShell.x86_64-linux = import ./devShell.nix { system = "x86_64-linux"; };
 


gytis-ivaskevicius:
So yeah, like how legit is using `darwinModules` key? As far as I am aware flakes don't support darwin as it is right now thus people tend to use `darwinModules` even tho it is 'unknown property', right?

Also if there is anything note-worthy in terms of darwin - I would love to hear it. This project has not really been tested on darwin :sweat_smile:

Enzime:
`darwinModules` is what `nix-darwin`'s flake itself is using:

https://github.com/LnL7/nix-darwin/blob/e1a3f7292f085fd588d11f94ed0f47968c16df0c/flake.nix#L22-L26

Flakes work fine on `darwin`, the only thing is just that `nix flake show` doesn't support `nix-darwin` but it doesn't support `home-manager` either, both `homeConfigurations`, `darwinConfigurations` and `darwinModules` all show up as `unknown`. I'm not sure if there are plans to fix this.

I've tested the `autoGenFromInputs` module on `darwin` and it works perfectly. The module should work on `home-manager` once my PR lands https://github.com/nix-community/home-manager/pull/2677

Enzime:
@gytis-ivaskevicius any thoughts on this PR?

gytis-ivaskevicius:
> nix-darwin but it doesn't support home-manager either, both homeConfigurations, darwinConfigurations and darwinModules all show up as unknown. I'm not sure if there are plans to fix this.

It is actually not possible to fix this. Nix itself does not have such functionality

gytis-ivaskevicius (approved):

Issue #113, opened 2022-02-27
Build fails with nix-darwin: The option `_module.args.inputs' is defined multiple times.


azuwis:
Demo `flake.nix` https://gist.github.com/azuwis/47a67c40f61afaa02362b79a0496afeb:

```
{
  description = "Darwin demo";

  inputs = {
    nixpkgs.url = github:nixos/nixpkgs/nixpkgs-unstable;
    utils.url = "github:gytis-ivaskevicius/flake-utils-plus/v1.3.1";

    darwin.url = "github:lnl7/nix-darwin/master";
    darwin.inputs.nixpkgs.follows = "nixpkgs";
  };

  outputs = inputs@{ self, nixpkgs, utils, darwin, ... }:
    utils.lib.mkFlake {
      inherit self inputs;

      hosts.demo.system = "aarch64-darwin";
      hosts.demo.output = "darwinConfigurations";
      hosts.demo.builder = args: darwin.lib.darwinSystem (removeAttrs args [ "system" ]);
    };
}
```

Error log:
```
$ nix build 'git+https://gist.github.com/azuwis/47a67c40f61afaa02362b79a0496afeb?ref=main#darwinConfigurations.demo.system'            
trace: warning: literalExample is deprecated, use literalExpression instead, or use literalDocBook for a non-Nix description.
trace: lib.crossLists is deprecated, use lib.cartesianProductOfSets instead
trace: warning: literalExample is deprecated, use literalExpression instead, or use literalDocBook for a non-Nix description.
trace: warning: lib.mkFixStrictness has no effect and will be removed. It returns its argument unmodified, so you can just remove any calls.
trace: `lib.nixpkgsVersion` is deprecated, use `lib.version` instead!
trace: warning: lib.readPathsFromFile is deprecated, use a list instead
trace: Warning: `showVal` is deprecated and will be removed in the next release, please use `traceSeqN`
trace: lib.zip is deprecated, use lib.zipAttrsWith instead
error: The option `_module.args.inputs' is defined multiple times.

       Definition values:
       - In `/nix/store/5m88sbxj815z8bzrij15bp4w4siakw1i-source/eval-config.nix'
       - In `<unknown-file>'
(use '--show-trace' to show detailed location information)
```

```
$ nix-info -m    
 - system: `"aarch64-darwin"`
 - host os: `Darwin 21.3.0, macOS 12.2.1`
 - multi-user?: `yes`
 - sandbox: `no`
[...]

PR #116, opened 2022-03-11
docs: nix-darwin builder workaround no longer necessary

montchr:
See discussion in https://github.com/gytis-ivaskevicius/flake-utils-plus/issues/113#issuecomment-1065546703


montchr:
On the other hand, it's possible that the upstream change is not yet available in the pinned input version in `flake.lock`?

fufexan (approved):

gytis-ivaskevicius:
Thank you!

PR #117, opened 2022-03-14
Hacky darwin fix

gytis-ivaskevicius:
Fix mentioned in https://github.com/gytis-ivaskevicius/flake-utils-plus/issues/113, need to play around with it and make sure that it does not break anything (but I think it does :/)


gytis-ivaskevicius:
Oh, tests are failing too. This is why I like TDD :sweat_smile:

PR #119, opened 2022-04-29
Hacky fix for darwin

bbuscarino:
For now, I just rebased onto latest master until I have time to mess with this more. I have long layovers for travel today, so I'll commit to debugging the failures then.

### DIFF
--- a/lib/mkFlake.nix
+++ b/lib/mkFlake.nix
@@ -161,7 +161,7 @@ let
                           config = selectedNixpkgs.config // config.nixpkgs.config;
                         } // { inherit (selectedNixpkgs) name input; };
                 }
-              else { _module.args.pkgs = selectedNixpkgs; })
+              else {})
 
               (optionalAttrs (options ? system.configurationRevision) {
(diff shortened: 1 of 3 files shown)


bbuscarino:
@gytis-ivaskevicius Finally got around to fixing this up, ready for review.

gytis-ivaskevicius:
Thanks <3

lib/mkFlake.nix
```diff
@@ -161,7 +161,7 @@ let
                           config = selectedNixpkgs.config // config.nixpkgs.config;
                         } // { inherit (selectedNixpkgs) name input; };
                 }
-              else { _module.args.pkgs = selectedNixpkgs; })
+              else { })
```
gytis-ivaskevicius, line 164: Is this change necessary? i feel like this would break something

gytis-ivaskevicius (approved):

lib/mkFlake.nix
bbuscarino: I don't remember why exactly this was necessary, I want to look into it later tonight. The tests were green though so...?

PR #120, opened 2022-05-31
use the channel lib

GTrunSec:
response this issue to https://github.com/divnix/digga/pull/469

### DIFF
--- a/lib/exportOverlays.nix
+++ b/lib/exportOverlays.nix
@@ -60,7 +60,8 @@ let
       fakePkgs = {
         callPackage = it: it;
         isFakePkgs = true; # Overlay maintainers may throw an exception in case this key is set in case overlay is not compatible with this function.
-        lib = flake-utils-plus.lib.internal;
+        # import the lib input from first channel
+        lib = inputs."${(builtins.elemAt (builtins.attrNames channels) 0)}".lib;
         system = "fake-system";
       };
 
--- a/lib/exportOverlays.nix
+++ b/lib/exportOverlays.nix
@@ -61,7 +61,7 @@ let
         callPackage = it: it;
         isFakePkgs = true; # Overlay maintainers may throw an exception in case this key is set in case overlay is not compatible with this function.
         # import the lib input from first channel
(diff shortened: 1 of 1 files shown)


GTrunSec:
We need a real channel to judge whether an overlay or isFunction.  https://github.com/gytis-ivaskevicius/flake-utils-plus/pull/120

lib/exportOverlays.nix
```diff
@@ -60,7 +60,8 @@ let
       fakePkgs = {
         callPackage = it: it;
         isFakePkgs = true; # Overlay maintainers may throw an exception in case this key is set in case overlay is not compatible with this function.
-        lib = flake-utils-plus.lib.internal;
+        # import the lib input from first channel
+        lib =  if (builtins.attrNames channels) == [] then flake-utils-plus.lib.internal else inputs."${(builtins.elemAt (builtins.attrNames channels) 0)}".lib;
```
gytis-ivaskevicius, line 64: How do we know what is first input? Also cant we simplify this as such:
```suggestion
        lib =  if (builtins.attrNames channels) == [] then flake-utils-plus.lib.internal else (builtins.elemAt (builtins.attrValues channels) 0).lib;
```

gytis-ivaskevicius (commented):
May I have the exact scenario where existing implementation fails? Or review from @blaggacao  or @Pacman99?

GTrunSec:
- https://github.com/gytis-ivaskevicius/flake-utils-plus/blob/master/lib/exportOverlays.nix#L80. 
- There is a problem with the way to judge an overlay function.

- tested by:  https://github.com/divnix/digga/pull/483/files#diff-15b26e32ee962cc4fbda5eea87245f72ed0b3710329e25e7a5c62a897610ce35R81

lib/exportOverlays.nix
GTrunSec: I don't know what you mean. IMO,  if we use an actual channel that should replace the flakePkg to https://github.com/divnix/digga/pull/483/files#diff-50b278a919472d6eab9c50638883df55c1129fb60f71ded49b06673428be8ceeR49

PR #134, opened 2023-05-13
🐛 Check all attributes for devshell legacyPackages

ravensiris:
Not sure if it adapts to the changes(https://github.com/numtide/devshell/pull/259) but at least it doesn't explode.
Fixes #133

PR #135, opened 2023-09-15
fix: check for system when looking for nixpkgs channels

jakehamilton:
`flake-utils-plus` currently has a bug to do with finding NixPkgs channels. It checks
for `legacyPackages.x86_64-linux.nix` on each flake input to determine if it is NixPkgs.
This typically works, but some flakes may define `legacyPackages` which can cause issues
if they omit the system used for this check (among other issues).
The fix adds an additional check to
ensure that the entire attribute path `legacyPackages.x86_64-linux.nix` exists rather
than expecting `x86_64-linux` to always be defined. This avoids the error that would prevent
integrating tools like [agenix](https://github.com/ryantm/agenix/tree/main) in a flake that
uses `flake-utils-plus`.

See a reproduction here: https://github.com/jakehamilton/fup-channels-bug

### DIFF
--- a/lib/mkFlake.nix
+++ b/lib/mkFlake.nix
@@ -205,7 +205,7 @@ mergeAny otherArguments (
 
         # Little hack, we make sure that `legacyPackages` contains `nix` to make sure that we are dealing with nixpkgs.
         # For some odd reason `devshell` contains `legacyPackages` out put as well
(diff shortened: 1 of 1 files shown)


jakehamilton:
Ah, didn't see that another PR fixes this same issue: https://github.com/gytis-ivaskevicius/flake-utils-plus/pull/134

gytis-ivaskevicius:
Thanks for the effort but I merged the other one

jakehamilton:
No worries @gytis-ivaskevicius, I hope you are feeling better!

PR #144, opened 2024-04-08
fix: Override option nixpkgs.config to avoid failed assertion

### LINKED ISSUE #142: Can't use flakes inputs modifying `nixpkgs.config` anymore
Hello, since the following lines [^1] were changed, it is not possible to have flake inputs with nixosModues modifying `nixpkgs.config`.
Including their modules causes an assertion error:

> Your system configures nixpkgs with an externally created instance.
> `nixpkgs.config` options should be passed when creating the instance instead.

I absolutely need this module, and I can't find a way to fix the issue.
Is there a way, either for the input flake to do things differently, or for me to fix something so that it works again?

[^1]: https://github.com/gytis-ivaskevicius/flake-utils-plus/blob/bfc53579db89de750b25b0c5e7af299e0c06d7d3/lib/mkFlake.nix#L149C18-L161C72

lordkekz:
Fixes #142 

Note that other modules could conceivably be affected if they read the value of `config.nixpkgs.config`, but I find that exceedingly unlikely.

### DIFF
--- a/lib/mkFlake.nix
+++ b/lib/mkFlake.nix
@@ -157,8 +157,9 @@ let
                         {
                           inherit (host) system;
                           overlays = selectedNixpkgs.overlays;
-                          config = selectedNixpkgs.config // config.nixpkgs.config;
+                          config = selectedNixpkgs.config // hostConfig.nixpkgs.config;
                         } // { inherit (selectedNixpkgs) name input; };
+                  nixpkgs.config = lib.mkForce { };
                 }
               else { })
 


gytis-ivaskevicius:
why mkforce is required there?

lordkekz:
Because we need to override the value that other modules have set.
Let's say `nixpkgs.config` is being set by some module `cheese.nix`
```nix
{ config, outputs, lib, ... }: {
  nixpkgs.config.allowUnfree = true;
}
```
And flake-utils-plus will add a module similar to this:
```nix
{ config, outputs, lib, ... }: {
  nixpkgs.pkgs = // generated from channels
}
```
Now, nixpkgs will throw an assertion failure because you can only set either `nixpkgs.config` or `nixpkgs.pkgs` but not both.
In the `nixpkgs.pkgs` which flake-utils-plus generates, the value `nixpkgs.config` is being applied, so we won't lose information by using the generated pkgs.

To get rid of the nixpkgs assertion failure, we need to reset `nixpkgs.config` to an empty set in the final evaluation; to override the `nixpkgs.config` which `cheese.nix` presumably sets without `mkDefault` or such, we need to choose a higher priority, like `mkForce`. If we used a low priority, it would have no effect, and if we used the regular priority, evaluation would fail because the conflict can't be resolved.

I think it's fine to use `mkForce` here because no user could ever want to override it; doing so would just break the configuration.

gytis-ivaskevicius:
Thanks!

PR #150, opened 2024-07-08
chore: Remove unused bindings

roberth:
Hi Gytis and others :wave:

I was wondering whether I could make `flake-utils-plus` load flake parts modules and/or vice versa, and found these unused bindings with https://github.com/oxalica/nil, so I figured I'd help out a tiny bit with this little cleanup first.

Flake parts is intended to be a mostly unopinionated common interface that helps with interoperability between pieces of Nix logic people write and publish for reuse. Users of `flake-utils-plus` benefit from this, as it makes arbitrary other flake-based functionality easy to import, so that they don't have to cobble up custom glue code to merge the flake outputs and then still be stuck with two frameworks that don't actually share things like packages very well.
I'll set it up so that it doesn't add a flake inputs dependency on flake-parts.

### DIFF
--- a/devShell.nix
+++ b/devShell.nix
@@ -18,7 +18,6 @@ let
   devshell = import devshellSrc { inherit system; };
 
   withCategory = category: attrset: attrset // { inherit category; };
-  util = withCategory "utils";
 
   rootDir = "$PRJ_ROOT";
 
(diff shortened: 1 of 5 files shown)


gytis-ivaskevicius (approved):

gytis-ivaskevicius:
Thank you <3

PR #156, opened 2026-01-17
Allow patching with functions that accept pkgs

PerchunPak:
This is useful for using `fetchpatch`:

```nix
channel.nixpkgs.patches = [
  (fetchpatch { ... })
];
```


tests/channel-patching-using-func/flake.nix
```diff
@@ -0,0 +1,82 @@
...
+
+          };
+      };
+
+
+    };
+}
+
+
+
```
PerchunPak, line 78: This is just a copy of `tests/channel-patching/flake.nix` so all questions to that file

gytis-ivaskevicius (approved):

